PR #70, opened 2026-06-03
Fix research preset classifier and extract web research into an extension

egorkhmelev:
Part 1 — classifier fix:
The web_research preset classifier always fell back to pro-search because it
passed the JSON schema with its $schema dialect URL, which the SDK's
structured-output validator rejects. Align classifyPreset with the proven
title-generator one-shot shape: strip $schema, use tools: [] + maxTurns: 2,
drop the unnecessary cwd, and handle error result subtypes. Add unit tests.

Part 2 — pluggable extensions:
Introduce a minimal code-extension system so web research becomes an optional
capability instead of core wiring. Extensions live in src/extensions/<name>/,
export a typed manifest (ArchieExtension), and contribute MCP servers +
PostToolUse hooks via a narrow ExtensionContext. extension-loader discovers
them by presence (with an optional ARCHIE_EXTENSIONS allowlist) and skips any
whose requiredEnv is unset. spawn.ts merges the combined contribution into
every agent track.

Move research-tools.ts into src/extensions/web-research/ and remove its direct
wiring from spawn.ts. The WebSearch/WebFetch denylist stays in core, so
removing the extension means "no web research", not "ungoverned web access".

Docs: add docs/architecture/extensions.md, update web-research / plugin-system
/ security / sdk-patterns references.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -92,7 +92,7 @@ Future work and unimplemented features. These are ideas that have been designed 
 - Entry point: `src/index.ts`
 - Agent spawners and registry: `src/agents/`
 - System orchestration: `src/system/`
-- MCP tools: `src/mcp/`
+- Web research (optional MCP tool + host hook): `src/extensions/web-research/` (wired in `src/agents/spawn.ts`)
 - Connectors (Slack, GitHub, OAuth, API): `src/connectors/`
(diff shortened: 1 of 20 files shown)


egorkhmelev:
Quite complicated rework, closing until we find better solution

PR #112, opened 2026-06-25
Pull plugin-declared repos in parallel on startup

egorkhmelev:
## What

`cloneRepos` (`src/system/workdir.ts`) previously awaited each plugin-declared repo's clone/fetch **one at a time**, so startup cost was the *sum* of every repo's git + network time. This pulls them all concurrently instead, so boot time now tracks the **slowest single repo** rather than the total.

```js
const results = await Promise.allSettled(
  repos.map(({ github, baseBranch }) =>
    cloneOrFetch(githubRepoToUrl(github), getBaseCachePath(github), github, baseBranch)
  )
);
results.forEach((result, i) => {
  if (result.status === 'rejected') {
    logger.warn('workdir', `Failed to clone/fetch ${repos[i].github}: ${result.reason}`);
  }
});
```

## Why it's safe to parallelize

- **Distinct target dirs.** Each repo clones to `getBaseCachePath(github)` (`workdir/repos/<org>/<repo>`), and the caller already dedupes by `github` — no two clones touch the same working tree or `.git`.
- **No cross-repo git lock.** Git's locks are per-repository; nothing operates on the same repo twice concurrently.
- **Shared-parent `mkdir` is safe.** Repos under the same org both call `mkdir(dirname, { recursive: true })` on the shared parent, which is idempotent and won't throw on `EEXIST`.

## Error handling

Switched to `Promise.allSettled` so one repo's failure logs a warning and boot continues, instead of aborting startup. This is a small improvement over the previous behaviour, where a failing fresh `git clone` would throw and crash boot (the fetch path already logged-and-continued).

## Why there's no race with task recovery

The call site stays `await`ed (`index.ts:123`), so the parallelism is internal to `cloneRepos`. Task recovery (`index.ts:250`) and webhook-driven agent spawns (`index.ts:253+`) still run only **after** every base cache is fully pulled — so `setupSharedClone`'s `git clone --shared` never reads a base cache mid-fetch. (Making the pull fire-and-forget *would* open that race; deliberately not doing that.)

## Not included

No concurrency cap — unnecessary for the current handful of repos. Worth revisiting only if the plugin-declared repo list grows into the dozens.

## Testing

- `npm run typecheck` ✅
- `npm run build` ✅

---
_Generated by [Claude Code](https://claude.ai/code/session_017rbc21ojtSKZUYmsq7Wm1d)_

### DIFF
--- a/src/system/workdir.ts
+++ b/src/system/workdir.ts
@@ -116,11 +116,21 @@ export async function bootstrapWorkdir(): Promise<void> {
 export async function cloneRepos(
   repos: Array<{ github: string; baseBranch?: string }>
 ): Promise<void> {
-  for (const { github, baseBranch } of repos) {
-    const repoPath = getBaseCachePath(github);
-    const repoUrl = githubRepoToUrl(github);
-    await cloneOrFetch(repoUrl, repoPath, github, baseBranch);
-  }
+  // Pull every repo in parallel. Each one targets its own directory, so there's
+  // no shared working tree or cross-repo git lock to contend over — total time
+  // becomes the slowest single repo instead of the sum of all of them.
+  // `allSettled` keeps one repo's failure from aborting startup (matching the
+  // "log and carry on" behaviour cloneOrFetch already has on the fetch path).
+  const results = await Promise.allSettled(
+    repos.map(({ github, baseBranch }) =>
+      cloneOrFetch(githubRepoToUrl(github), getBaseCachePath(github), github, baseBranch)
+    )
+  );
+  results.forEach((result, i) => {
+    if (result.status === 'rejected') {
+      logger.warn('workdir', `Failed to clone/fetch ${repos[i].github}: ${result.reason}`);
+    }
+  });
 }
 
 /**


egorkhmelev:
## E2E verification (live boot from this branch)

Rebased onto current `main` (clean, no conflicts) and verified against a live instance booted from this branch via the `archie-e2e` harness, using the real plugins config — 5 declared repos: `sweatco/sweatcoin-backend`, `sweatco/sweatcoin-infrastructure`, `sweatco/sweatcoin-mobile`, `sweatco/archie-plugins`, `sweatco/archie-hq`. Instance attested as composed from the checkout under test and reached healthy.

### Parallel dispatch confirmed

All 5 pulls started within **13 ms** of each other:

```
11:34:19.826  [System] Pulling latest for sweatco/archie-plugins...
11:34:19.830  [System] Pulling latest for sweatco/archie-hq...
11:34:19.832  [System] Pulling latest for sweatco/sweatcoin-backend...
11:34:19.834  [System] Pulling latest for sweatco/sweatcoin-infrastructure...
11:34:19.839  [System] Pulling latest for sweatco/sweatcoin-mobile...
```

Under the previous sequential loop, repo N+1's line could only be printed after repo N's four git commands (`remote prune`, `fetch --all`, `checkout`, `reset --hard`) had all completed — so the flat 13 ms spread is direct evidence of the fan-out.

### A/B against the sequential version, identical warm caches

Measured the pull phase (first `Pulling latest` → `Plugins loaded`) on the same instance and the same warm base caches, patching `cloneRepos` back to the sequential loop in between and then reverting it:

| variant | pull phase | start spread |
|---|---|---|
| parallel (this PR) | **7.55s** | 13 ms |
| sequential (`main`) | **12.84s** | staggered by 2.2s / 1.3s / 3.2s / 2.5s |

~1.7× on warm caches. The gap widens on colder caches, where `sweatcoin-mobile` (1.2 GB) dominates and the sequential total is the sum rather than the max.

### Correctness of the concurrent writes

This was the actual risk in the change — five concurrent `git reset --hard` runs. After the parallel boot, every base cache was on its declared base branch, clean, and error-free:

```
archie-plugins            branch=main    dirty=no
archie-hq                 branch=main    dirty=no
sweatcoin-backend         branch=master  dirty=no
sweatcoin-infrastructure  branch=master  dirty=no
sweatcoin-mobile          branch=main    dirty=no
```

Zero `warn`/`error`/`failed` lines in the boot log.

### Failure isolation (`Promise.allSettled`)

Injected a sixth, deliberately nonexistent repo declaration to force a fresh-clone rejection. Boot survived and completed:

```
[System] Cloning sweatco/archie-does-not-exist-pr112 from https://github.com/sweatco/archie-does-not-exist-pr112.git...
[workdir] Failed to clone/fetch sweatco/archie-does-not-exist-pr112: Error: Command failed: git clone -b "main" ...
          fatal: repository 'https://github.com/sweatco/archie-does-not-exist-pr112.git/' not found
Plugins loaded: archie, data-analytics, engineering, marketing, ops, pm, qa
```

The warning named the correct repo, which confirms the `results[i]` ↔ `repos[i]` index alignment that `Promise.allSettled`'s order guarantee relies on. The other five repos pulled normally and the instance went healthy.

### Also green

`npm run typecheck`, `npm run build`, and `npm test` (63 files, 832 tests) all pass on the rebased tree. Teardown was clean — no containers left behind.

### What was not verified

- **All five repos cloning fresh at once.** The run above was five fetches plus one fresh clone. Bandwidth/disk contention for ~2.2 GB of concurrent fresh clones is unmeasured; it affects first-boot-on-an-empty-workdir only.
- **The old crash-on-fresh-clone-failure behaviour** described above was not reproduced directly — only the new log-and-continue behaviour was observed.
- **No concurrency cap** at this repo count, as the description notes. Five is fine; dozens remains the revisit point.

### One thing worth deciding separately

Because a failed repo no longer aborts boot, the agent that declared it still registers with it (`also mounts: sweatco/archie-does-not-exist-pr112`). So a typo'd repo in plugin config now boots "successfully" with an agent that will fail later at task time, rather than failing loudly at startup. That is the deliberate trade-off here, not a defect in this change — but it may deserve a follow-up (skip registration when the base cache is missing, or escalate the log level).

PR #126, opened 2026-06-28
Make agent background-task waits work: re-engage, no spurious recovery, CLI visibility

egorkhmelev:
## Problem

A repo agent backgrounds a wait (`Bash` with `run_in_background`, e.g. `sleep`/CI poll) and ends its turn expecting to be re-invoked when it settles. Interactive Claude Code's REPL does that on `task_notification`; **archie doesn't** — agents are driven solely by their MessageQueue and only reacted to `system:init`/`result`, discarding the task events. So the settled task woke nothing: the agent stalled, and recovery fired *during* the wait (all agents looked idle), escalating toward a nuclear restart.

## Fix

**Re-engagement + recovery**
- Track in-flight background tasks per agent (`Agent.backgroundTasks`): added on SDK `task_started`, drained on `task_notification` and on subprocess exit.
- On settle, enqueue a wake message so the agent resumes and re-checks — archie supplies the re-invocation the SDK expects the harness to provide.
- `idleDecision` treats an agent with in-flight background tasks as **busy**, so recovery/quiescence holds off during a legitimate wait.
- ✅ **Confirmed live**: the SDK *does* emit `task_started`/`task_notification` on the streaming output in archie's headless setup; auto-resume works with no recovery during the wait.

**Visibility**
- New `agent:bg_task` event → the CLI renders **one entry per task**, ⏳ running folded to ✅/❌ on settle (transcript re-derives on each event). Server logs gain per-agent start/settle lines.
- The agent is shown **idle** during the wait — it isn't actively working; the ⏳ entry is the accurate "in progress" signal (we briefly faked "active", then reverted it as misleading + redundant).

**CLI event-drop fix (general, surfaced here)**
- The CLI fed live SSE events to `TaskDetail` one at a time via a single prop (replace semantics), so a burst collapsed under React batching — an inter-agent message immediately followed by `agent:active` in the same millisecond was silently dropped from the transcript (a `mobile→pm` report vanished from the CLI while present in logs + events.jsonl). App now accumulates events in an **append-only queue**; `TaskDetail` processes the delta. Fixes any event burst, not just background tasks.

## Files
- `src/agents/agent.ts`, `src/agents/spawn.ts`, `src/tasks/recovery.ts` — tracking, re-engage on settle, busy-aware `idleDecision`, server log lines
- `src/system/event-bus.ts`, `src/cli/components/TaskDetail.tsx`, `src/cli/App.tsx` — `agent:bg_task` ⏳→✅ entry + event-queue fix

## Testing
- Unit: `idleDecision` busy-on-background-task case. **476/476** green, typecheck clean.
- Smoke (confirmed): background a `sleep` → no `[recovery]` during the wait, agent auto-resumes on completion, ⏳→✅ entry in the CLI, agent shows idle during the wait, and the inter-agent completion message renders (no longer dropped).

Builds on the completion-as-quiescence model (#125, merged) — same recovery/quiescence path.

### DIFF
--- a/src/agents/agent.ts
+++ b/src/agents/agent.ts
@@ -35,6 +35,16 @@ export class Agent {
   sandbox?: SandboxOptions;
 
   /**
+   * In-flight background task ids (SDK `task_started` → `task_notification`): a
+   * backgrounded Bash wait, a subagent, etc. The agent's turn can END while one
+   * is still running, so the idle-check treats an agent with non-empty
+   * backgroundTasks as *busy* (not stalled) — otherwise recovery fires under a
+   * legitimate wait. Drained on settle (the spawn loop re-engages the agent) and
+   * on subprocess exit (a dead process can't settle them).
+   */
+  readonly backgroundTasks: Set<string> = new Set();
+
+  /**
    * A task teardown (complete/stop) deferred until this agent's current turn
    * ends. The spawn loop runs it when it sees the SDK `result` event, so the
    * triggering tool's response and the Stop hook's control round-trip both
--- a/src/agents/spawn.ts
+++ b/src/agents/spawn.ts
@@ -653,8 +653,10 @@ Shared folder: ${sharedPath} [READ-ONLY]
             // not stalled — no spurious recovery) and re-engage the agent on settle.
             if (event.type === 'system' && event.subtype === 'task_started') {
               agent.backgroundTasks.add(event.task_id);
+              logger.agent(def.id, `background task started — ${event.description}`);
             } else if (event.type === 'system' && event.subtype === 'task_notification') {
               agent.backgroundTasks.delete(event.task_id);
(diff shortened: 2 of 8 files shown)


egorkhmelev:
Added a second commit (`3b6f83f`): **visibility** — an agent with in-flight background work now stays shown as `active` (Slack "Archie is …" indicator + agent:active state), instead of going idle the moment it backgrounds a wait. `updateAgentState` suppresses the turn-end inactive transition while `backgroundTasks` is non-empty; it flips inactive when the task settles and the resumed turn finishes, or on subprocess exit. Plus per-agent CLI log lines on background-task start/settle.

Smoke check for this commit: background a `sleep` → the agent stays active in the indicator during the wait, CLI shows `background task started — …` then `background task completed — …`, and it auto-resumes.

PR #135, opened 2026-06-29
Memory v2: simplify recovery, visibility, and housekeeping

hardworker:
## Summary

Third PR in the Memory v2 stack. This is the whole-branch simplification and adversarial hardening pass.

- Makes extraction intent, crash replay, and task completion durable.
- Serializes visibility/transcript snapshots so private Slack content cannot race into public memory.
- Unifies Slack identity handling and fail-closed privileged actions.
- Derives trigger destination visibility live while preserving prompt-origin confidentiality.
- Replaces model-driven housekeeping with deterministic validation, deduplication, staleness, archival, and alias reconciliation.
- Centralizes escaped evidence envelopes for all persisted memory reads and injection.
- Removes dormant/test-only APIs, repeated transcript parsing, timing-based tests, and excessive implementation commentary.
- Aligns architecture, security, orchestration, trigger, persistence, and rollout documentation.

## User and developer impact

- Triggers stay enabled when a privacy lookup or public-crossing check defers a run; quota and schedule conditions are unchanged until a real fire.
- Public tasks downgrade durably before private messages or edits are written.
- Unknown Slack identities remain retryable; known-untrusted automation is redacted without wedging the thread.
- Completion does not become irreversible until the extraction intent is on disk.
- Memory tools return bounded historical evidence rather than instruction-shaped persisted text.

## Stack

1. #276 — bounded injection and selection telemetry
2. #277 — read tools, confidentiality, and collaboration profiles
3. **This PR — simplification, recovery, and visibility hardening**
4. #228 — evaluation harness

## Review

Three adversarial fix/re-review iterations resolved all confirmed high- and medium-severity findings. One low product-policy question remains intentionally open: how to reduce fresh collaboration-profile facts above a soft cap without silently hiding valid preferences.

## Verification

- `npm run typecheck`
- `npx vitest run --testTimeout=10000` — 1,208 tests across 86 files
- `git diff --check`


src/system/trigger-scheduler.ts
```diff
@@ -360,7 +360,7 @@ export async function fireTrigger(trigger: Trigger, context: FireContext): Promi
     return;
   }
 
-  const task = await Task.create();
+  const task = await Task.create('public');
```
egorkhmelev: **Blocking.** `Task.create('public')` is unconditional, so a `channel_message` trigger bound to a private channel produces a public task with that private thread linked — and the extraction gate at `src/memory/lifecycle.ts:104` then passes, writing private content into the org-public store. The ingest path derives visibility correctly (`Task.create(thread.taskVisibility)`, `src/connectors/slack/events.ts:773`); this path does not.

Reproduced with a test driving the real `fireTrigger`, `Task.create/get/append`, `findTaskByThread`, and memory lifecycle: 5/5 passing, leaking entity, activity, summary, and profile artifacts, with a control confirming the ingest path skips the same channel.

Also note `propose_trigger` is registered unconditionally (`src/agents/tools.ts:2649`), so a *private* task can mint a trigger whose fires are public — a confidentiality escalation. Trigger creation validates no channel privacy anywhere (`buildConditions`, `handleTriggerApproval`); `triggerVisibleFrom` governs listing only.

src/memory/context.ts
```diff
@@ -52,45 +62,130 @@ export async function buildMemoryContext(
...
+ * alters the prompt; skipped without a `taskId` and when injection is off, so
+ * the collect-only posture stays write-free.
+ */
+async function recordSelection(
+  selectors: MemorySelectors,
+  users: UserRef[],
+  selection: SelectionResult | null,
+  context: string,
+): Promise<void> {
+  if (!selectors.taskId || !isInjectionEnabled()) return;
```
egorkhmelev, line 115: **Blocking.** This gates on `taskId && isInjectionEnabled()` only — not on task visibility. A task whose `metadata.json` says `visibility: "private"` therefore writes `memory/tasks/<id>/telemetry.jsonl` containing the DM-derived task title, participant Slack ids, and display names. Reproduced with raw-byte assertions on all three.

Latent behind the default-off flags (control confirms nothing is written with shipped defaults), so this is at-rest exposure that appears when the documented rollout flips them. Either gate on visibility here, or move telemetry outside the `memory/` tree.

src/memory/telemetry.ts
```diff
@@ -0,0 +1,99 @@
...
+ * they are agent-authored queries, not secrets.
+ */
+export async function recordPull(
+  taskId: string | undefined,
+  agent: string | undefined,
+  toolName: string,
+  args: Record<string, unknown>,
+  result: PullRecordResult,
+): Promise<void> {
+  if (!taskId) return;
```
egorkhmelev, line 69: **Blocking (same defect as `context.ts:110`).** `recordPull` has no visibility check at all, so with tools enabled a private task persists the agent's verbatim `search_memory` query under `memory/`. Verified present in the reproduced `telemetry.jsonl`.

src/memory/lifecycle.ts
```diff
@@ -271,7 +386,44 @@ export function resolveFallbackId(metadata: TaskMetadata): UserRef {
...
+  for (const name of entries) {
+    if (!name.endsWith('.md')) continue;
+    let dest: string;
+    try {
+      dest = getSummaryPath(name.slice(0, -3)); // the real path guard decides validity
+    } catch {
+      continue;
+    }
+    await mkdir(dirname(dest), { recursive: true });
+    await rename(join(legacyDir, name), dest);
```
egorkhmelev: **Blocking.** `mkdir` + `rename` with no try/catch, reached from `initMemory()` which `src/index.ts:111` awaits inside the startup try whose catch calls `process.exit(1)`. Runs on every startup, so it is on the deploy path for every existing store.

Both consequences reproduced (4/4 passing): a single `EISDIR` or `EACCES` on one legacy summary rejects `initMemory()` and exits the service instead of degrading; and POSIX `rename` overwrites silently, replacing an existing `summary.md` with stale legacy content while `logger.warn`/`logger.error` are never called and the run still logs `migrated 1 task summaries`.

Confirmed a regression, not pre-existing: on `main` the equivalent step is a recursive `mkdir`, which returns without throwing on an existing-but-unwritable directory. Worth guarding the unguarded `mkdir` at `src/memory/index.ts:38` in the same pass.

src/memory/lifecycle.ts
```diff
@@ -271,7 +386,44 @@ export function resolveFallbackId(metadata: TaskMetadata): UserRef {
...
 
 /**
- * Write the per-task summary to workdir/memory/summaries/<taskId>.md.
+ * One-time layout migration: memory/summaries/<taskId>.md →
+ * memory/tasks/<taskId>/summary.md. Idempotent — a missing legacy dir is a
+ * no-op, non-matching files are left in place, and the legacy dir is removed
+ * only once emptied.
+ */
+export async function migrateLegacySummaries(): Promise<void> {
+  const legacyDir = join(getMemoryDir(), 'summaries');
```
egorkhmelev: **Blocking.** This migration exists to carry a pre-visibility store forward, but everything it moves was written by `main`, which had no visibility gate — verified: at merge-base `51e876d` this file has zero matches for `visibil`, and `isMemoryEnabled()` was already default-on. So DM-derived summaries land at `memory/tasks/<id>/summary.md`, indistinguishable from new ones.

Reproduced end to end: migration promotes DM content, `read_task_summary` returns it verbatim to an unrelated task, `search_memory` surfaces it plus the never-migrated `recent-activity.md` row, and `buildMemoryContext` injects that row. This makes `docs/architecture/memory.md`'s "Recent activity contains only public-task output" false for any pre-existing store.

Code and contract disagree on whether legacy stores can exist. Options: filter/exclude legacy artifacts here; drop the migration; or warn loudly / refuse, matching the `warnLegacyUserFiles()` precedent.

src/tasks/persistence.ts
```diff
@@ -203,12 +203,12 @@ export async function loadMetadata(taskId: string): Promise<TaskMetadata | null>
...
 
-/**
- * Format a log entry for the shared knowledge log
- */
+/** Format a log entry so body continuations cannot mimic source lines. */
 function formatLogEntry(entry: LogEntry): string {
   const typeStr = entry.type ? ` [${entry.type}]` : '';
-  return `[${entry.timestamp}] [${entry.source}]${typeStr} ${entry.message}\n`;
+  const safeSource = entry.source.replace(/[\r\n]+/g, ' ');
+  const framedMessage = entry.message.replace(/\r\n?/g, '\n').replace(/\n/g, '\n  ');
```
egorkhmelev, line 235: This re-indent is the only thing making the new ownership check unspoofable, since `AUTHOR_LINE_RE` and `MSG_ID_LINE_RE` are line-anchored — an un-indented body line forges both an author and an evidence id.

It has no producer-side test. Mutation-tested: deleting `.replace(/\n/g, '\n  ')` still leaves `src/tasks/__tests__/persistence.test.ts` passing 13/13. The test that claims to cover it (`src/memory/__tests__/lifecycle.test.ts:853`) hand-writes an already-indented line, so it cannot detect the regression. Worth pushing a crafted multi-line message through the real append path and asserting `extractAuthorUsers` yields only the true author.

src/memory/store.ts
```diff
@@ -30,36 +30,81 @@ export async function readUser(username: string): Promise<string> {
...
+  for (const id of new Set(userIds)) {
+    let text: string;
+    try {
+      text = await readUser(id);
+    } catch (error) {
+      logger.warn('memory', `readUserFiles: skipping invalid or unreadable user ${JSON.stringify(id)}: ${error}`);
+      continue;
+    }
+    if (!text) continue;
+    const displayName = parseUserDisplayName(text);
```
egorkhmelev: `parseUserDisplayName`'s double-quote-only regex makes `readUserFiles` skip any profile whose `display_name` is absent, unquoted, or single-quoted — while the push path (`src/memory/context.ts:59` `readUser`) injects the same file with no such requirement. Verified 4/5 failing with `push=true, pull=false`: the agent is told a fact it cannot then look up.

Reachable rather than synthetic — the earliest `store.ts` (`2e1a83c`) wrote user files with no frontmatter at all.

src/connectors/slack/events.ts
```diff
@@ -71,6 +71,31 @@ export interface SlackLifecycle {
...
+    logger.warn('Slack', `Ignoring ${actionName}: action has no actor`);
+    return null;
+  }
+  try {
+    const info = await getUserInfo(userId);
+    if (isExternalUser(info)) {
+      logger.system(`Ignoring ${actionName} from external/guest user ${userId}`);
+      return null;
+    }
+    return { id: userId, info };
```
egorkhmelev, line 145: Failing closed on actor classification is right; failing *silently* is the problem. On any `users.info` error this returns null and all seven handlers return straight after `ack()` — no `updateMessage`, no ephemeral, no log the clicker can see. A transient rate-limit means an internal engineer clicks Approve, nothing visibly happens, and the task stays paused.

The new test asserts exactly this silence, so it reads as intended. An ephemeral "couldn't verify you, try again" would preserve the guard and restore the signal.

src/memory/tools.ts
```diff
@@ -0,0 +1,326 @@
...
+  const readEntityTool = tool(
+    'read_entity',
+    'Read one entity page in full (facts, relations) by slug — slugs come from the entity index or search_memory results. Aliases resolve to their canonical page.',
+    {
+      slug: z.string().describe('Entity slug (lowercase-kebab) or a known alias'),
+    },
+    async (args) => {
+      const raw = args.slug?.trim() ?? '';
+      // Reject path-capable aliases before any filesystem access.
+      const safeAliasShape = /^[A-Za-z0-9 _-]{1,64}$/.test(raw);
```
egorkhmelev: This hand-rolled `safeAliasShape` regex sits next to the `isValidEntitySlug` call, contradicting the invariant stated in this file's header and in `src/memory/CLAUDE.md:26` ("Every identifier passes `paths.ts` guards").

Verified safe today — `../secrets` is rejected with `isError` and zero fs calls across 27 passing cases — so this is a contract defect rather than a live escape. Worth moving beside `isValidEntitySlug` so a future change that turns an alias into a path inherits the guard instead of the rule being true in the doc and false in the code.

src/memory/tools.ts
```diff
@@ -0,0 +1,326 @@
...
+  return hits
+    .sort(
+      (a, b) =>
+        b.score - a.score || KIND_RANK[a.kind] - KIND_RANK[b.kind] || a.id.localeCompare(b.id),
+    )
+    .slice(0, maxHits);
+}
+
+/** Read every public task summary. */
+async function readAllSummaries(): Promise<Array<{ taskId: string; text: string }>> {
```
egorkhmelev: `search_memory` does an unbounded, uncached full-store scan per call: every entity page plus a `readdir` of `memory/tasks/` and a `readFile` of every `summary.md`. Task summaries are never pruned — only `recent-activity.md` is trimmed and only entity *count* is capped — and once the flags flip, `appendTelemetry` creates a task directory for every task, public or private, so the denominator grows per task rather than per summarized task.

Agents are told to call this freely, and nothing bounds the input the way `SEARCH_MAX_HITS`/`RESULT_MAX_CHARS` bound the output. Consider indexing summaries or reusing the bounded activity index.

src/connectors/slack/client.ts
```diff
@@ -1181,52 +1181,63 @@ export function isExternalUser(user: {
...
+  );
+}
+
+/**
+ * Returns whether a channel is shared with one or more external Slack
+ * workspaces (Slack Connect). D-prefixed ids are still checked because Slack
+ * Connect DMs may be externally shared. Result is cached for
+ * 1 minute. On API failure, returns false (fail-open — this is advisory only).
+ */
+export async function isChannelShared(channelId: string): Promise<boolean> {
```
egorkhmelev, line 1703: This refactor's stated benefit doesn't materialize: the header claims one `conversations.info` snapshot per channel, but `getChannelInfo` at `:1274` still makes its own uncached call, so `fetchSlackThread` issues two per event exactly as before.

The one real behavior change is dropping the `channelId.startsWith('D')` short-circuit, which (a) adds a `conversations.info` call for every DM event that previously made none, and (b) lets a `D` conversation report `shared: true` into `sendSharedChannelWarnings`, stamping `isShared` and posting the external-org ephemeral inside a 1:1. Since the visibility feature uses `getChannelInfo` rather than this function, leaving `isChannelShared` alone looks equivalent and smaller.

src/memory/activity.ts
```diff
@@ -17,11 +17,11 @@ const HEADER = `# Recent Activity
 | Date | Task ID | Summary | Domain | User |
 |------|---------|---------|--------|------|`;
 
-const ROW_REGEX = /^\|\s*([^|]+)\s*\|\s*([^|]+)\s*\|\s*([^|]+)\s*\|\s*([^|]+)\s*\|\s*([^|]+)\s*\|$/;
+const ROW_REGEX_5 = /^\|\s*([^|]+)\s*\|\s*([^|]+)\s*\|\s*([^|]+)\s*\|\s*([^|]+)\s*\|\s*([^|]+)\s*\|$/;
```
egorkhmelev: `readActivity()` → `renderActivityTable()` parses and rebuilds an identical table with no filtering, transformation, or visibility check in between, where `main` injected the file's bytes directly. The only behavior delta is that rows failing the 5-column regex (a summary containing `|`, a hand-edited row, a legacy 4-column row) are now silently dropped from the injected block.

`ROW_REGEX` → `ROW_REGEX_5` also implies a second arity that doesn't exist — a leftover from the discarded 6-column `access:` design. Either keep the direct read or apply the filtering that would justify the round trip.

docs/plans/README.md
```diff
@@ -23,6 +23,7 @@ Historical record of Archie's incremental development. Each plan represents a mi
 | [v10](v10-agent-recovery-impl.md) | Agent recovery implementation | Partially implemented |
 | … | (v11–v31 — see files in this directory) | — |
 | [20260626](20260626-response-formatting.md) | Message footer + self-updating PR cards | Implemented |
+| [20260719](20260719-memory-v2.md) | Memory v2: bounded injection, pull tools + eval harness, confidentiality layer | Implemented (rollout gates open) |
```
egorkhmelev: This row claims "pull tools + eval harness", but `docs/proposals/memory-v2-roadmap.md` and `docs/plans/20260719-memory-v2.md` in this same diff both state the harness was split out to #228 and is not on this branch. Since `docs/architecture/memory.md` is being installed as the single as-built source, this is the one line that misreports what merged — and the first thing a reader scanning plan history hits.

egorkhmelev (changes requested):
# Review: Memory v2 — bounded injection, pull tools, confidentiality layer

**Event:** `REQUEST_CHANGES` · **Reviewed at:** `bcb6e86` · Graded on the assumption this lands **dark** (both flags false, memory recall not expected to improve). If the goal is to enable memory here, more of the non-blocking list moves up.

Good change overall — dropping the `scope: org` exemption and `SCORE_ORG` is the right fix for the 381K-token blowup, the always-injected index covers what selection misses, and one immutable task-level `visibility` is a simpler model than the ACL prototype it replaces. Verified working, each with a control: the extraction gate fails closed on private/absent/unrecognized visibility, `search_memory` excludes non-authors even as the strongest textual match, and both flags ship false.

Four blockers below, each reproduced by a test driving real code paths. Two are independent of flag state — they bite on the next deploy whether memory is ever enabled.

## Blocking

**1. Trigger tasks are stamped `public` regardless of channel privacy** — `src/system/trigger-scheduler.ts:363`. `fireTrigger` hardcodes `Task.create('public')` where the ingest path uses `Task.create(thread.taskVisibility)`, and nothing at trigger creation or approval checks channel privacy. A trigger on a private channel therefore produces a public task with that private thread linked, and extraction writes the content to entities, activity, summary, and profiles. Reproduced 5/5 against real `fireTrigger`/`Task`/`findTaskByThread`/lifecycle, with a control confirming ingest correctly skips the same channel. Worse: `propose_trigger` is registered unconditionally, so a *private* task can mint a trigger whose fires are public — a confidentiality escalation. Your two new docs conflict here (`memory.md:13` vs `:14`); please decide whether trigger tasks are legitimately public, and either say so in the docs or derive visibility as ingest does.

**2. Private tasks write private-derived data into `memory/`** — `src/memory/context.ts:110`, `src/memory/telemetry.ts:66`. `recordSelection` gates on `taskId && isInjectionEnabled()` only; `recordPull` has no visibility check. Reproduced against a task whose `metadata.json` says `visibility: "private"`: `telemetry.jsonl` holds the DM-derived title, Slack id, display name, and the agent's verbatim query. Latent behind the default-off flags (control confirms), so it lands when the documented rollout flips them. Gate both sensors, or move telemetry out of the `memory/` tree.

**3. `migrateLegacySummaries` is unguarded** — `src/memory/lifecycle.ts:411-412`. No try/catch, reached from `initMemory()` inside the startup try whose catch is `process.exit(1)`, and it runs on every boot. Reproduced 4/4: one `EISDIR`/`EACCES` exits the service instead of degrading, and `rename` silently replaces an existing `summary.md` with no warn/error while still logging `migrated 1 task summaries`. A regression, not pre-existing — main's recursive `mkdir` returns fine in exactly the state where the new `rename` dies. Worth guarding `src/memory/index.ts:38` in the same pass.

**4. Legacy carry-forward contradicts "public by construction"** — `src/memory/lifecycle.ts:395`. Verified at merge-base `51e876d`: zero matches for `visibil`, `isMemoryEnabled()` default-on — so any instance that ran main has DM-derived summaries and activity rows on disk. Reproduced: migration promotes DM content, `read_task_summary` returns it verbatim to an unrelated task, `search_memory` surfaces it plus the never-migrated activity row, and `buildMemoryContext` injects that row — making `memory.md`'s "Recent activity contains only public-task output" false for any existing store. Filter legacy artifacts, drop the migration, or warn/refuse per the `warnLegacyUserFiles()` precedent.

## Non-blocking

- **`src/tasks/persistence.ts:210`** — the body re-indent is the only thing making the ownership check unspoofable, and it has no producer-side test: deleting it leaves `persistence.test.ts` green at 13/13. The test that claims to cover it hand-writes an already-indented line.
- **`src/memory/store.ts:72` vs `context.ts:59`** — double-quote-only `display_name` regex means push injects a legacy profile that pull can't find (4/5 failing). Reachable: the earliest `store.ts` wrote no frontmatter at all.
- **`src/memory/paths.ts:160`** — `SLACK_ID_RE` accepts `B…`/`T…`, so an internal bot can earn a profile and enter pull-tool author scope.
- **`src/connectors/slack/events.ts:92`** — a transient `users.info` failure drops an approval *silently* (no update, no ephemeral, no log). Failing closed is right; an ephemeral would restore the signal.
- **`src/memory/tools.ts:184`** — `search_memory` is an unbounded uncached full-store scan per call, and once flags flip every task gets a directory, not just summarized ones.
- **`src/memory/tools.ts:270`** — hand-rolled alias regex contradicts the "every identifier passes `paths.ts` guards" invariant. Traversal verified safe today (27 cases, `../secrets` rejected, zero fs calls), so contract defect only.
- **Smaller:** stale comment at `src/agents/tools.ts:2189` now says the opposite of the code; the `isChannelShared` refactor doesn't achieve its stated caching benefit but does add a `conversations.info` call per DM and can stamp `isShared` in a 1:1; `activity.ts:20`'s parse/re-render round trip only adds silent row loss; `extractUsernames`/`extractRequestingUser` are now dead; several `context.ts`/`entities.ts` exports have no in-tree consumer; `docs/plans/README.md:26` claims the eval harness ships here when two other docs say otherwise.

## Questions on scope

1. Split the seven-handler internal-actor guard into its own PR against `main`? No memory dependency, and #176 set that precedent — bundled here it can't land until the whole stack does.
2. Is retiring the memory layer's "no type changes propagate" ejectability clause intended? `visibility` is now required on `TaskMetadata` with three non-memory call sites, derived inside the Slack client.
3. Is deleting `openspec/specs/memory-layer/spec.md` intended rather than freezing it? 579 lines, no replacement, four archived changes still cite the path, two committed requirements reversed without being recorded as reversals.
4. Should the snapshot-and-clear precondition move into `docs/guides/deployment.md`? Verified untouched; the step is irreversible forward.

## Verification

Isolated worktree at `bcb6e86`: `npm ci` clean, typecheck zero diagnostics, suite **919/919** across 66 files, memory **381/381**. Note the PR cites 922/922 — memory matches exactly at 381, so that total looks like a different base. Deps are pre-bump (SDK 0.3.202, MCP 1.29.0) vs main's `^0.3.220`/`^1.30.0`.

Scope, stated so it isn't over-read: the four blockers were each reproduced with controls and three ACs positively verified; the rest were exercised but their evidence didn't meet this review's citation bar, so they're unattested rather than passing. No live E2E boot, and the two enablement ACs depending on #228's harness can't be closed from this branch.

src/system/trigger-scheduler.ts
hardworker: Fixed in `1d2552f`. Trigger fires no longer default to public: message-trigger tasks inherit the source event visibility, scheduled channel triggers resolve the channel's current privacy, and DM delivery is private. Private tasks cannot approve persistent triggers, and legacy triggers without a verified public creation context are paused. Added end-to-end regression coverage for private message-trigger fires.

src/memory/context.ts
hardworker: Addressed in `1d2552f` with a deliberate variant of the suggested options. We retain full telemetry for both public and private tasks, but moved it out of the agent-readable task corpus to `memory/telemetry/tasks/<taskId>/telemetry.jsonl`. Prompt assembly and memory tools only read the public corpus paths; a regression test seeds a private query in telemetry and verifies `search_memory` cannot surface it. Keeping the subtree under `memory/` lets operators download/snapshot the corpus and telemetry in one batch; the deployment docs mark telemetry as sensitive operator data.

src/memory/telemetry.ts
hardworker: Same fix as the selection sensor in `1d2552f`: private pull records and verbatim queries are intentionally retained under `memory/telemetry/tasks/`, tagged with task visibility, but that subtree is not reachable through prompt injection or any memory tool. Tests verify a private query is recorded there and remains absent from search results.

src/memory/lifecycle.ts
hardworker: Fixed in `1d2552f`. `migrateLegacySummaries` was removed. Store preparation is now guarded inside `initMemory()`; initialization errors disable all memory reads/writes for the process and log the reason instead of escaping into the application startup path. Regression coverage includes an invalid memory path and verifies that no lifecycle handler is registered.

src/memory/lifecycle.ts
hardworker: Fixed in `1d2552f` by refusing carry-forward entirely. A fresh store gets `.public-store-v1`; an unmarked non-empty store is left untouched and memory is disabled for the process. The deployment guide now requires stopping Archie, snapshotting `memory/`, and clearing it before the first deployment with this boundary.

src/tasks/persistence.ts
hardworker: Added the requested producer-side regression test in `1d2552f`. It sends a crafted multiline body through real `appendSlackMessage`, verifies the continuation line is indented, and asserts `extractAuthorUsers` returns only the true Slack author rather than the forged author/evidence line.

src/memory/store.ts
hardworker: Fixed in `1d2552f`. `parseUserDisplayName` now accepts generated double-quoted YAML plus legacy single-quoted and unquoted values, and `readUserFiles` falls back to the stable Slack ID when the field is absent. The store tests cover all legacy shapes.

src/connectors/slack/events.ts
hardworker: Fixed in `1d2552f`. Actor lookup failures still deny the action, but now log the classification failure and post an ephemeral ‘couldn't verify your Slack account; please try again’ message to the actor in the originating thread. Added shared tests across the interactive authorization handlers.

src/memory/tools.ts
hardworker: Fixed in `1d2552f`. The alias-shape boundary moved to `paths.ts` as `isValidEntityLookup`; `read_entity` now uses that shared guard before filesystem access. Path tests cover guarded aliases and traversal-shaped input.

src/memory/tools.ts
hardworker: Fixed in `1d2552f` by removing the full `memory/tasks/` scan from `search_memory`. Search now reads the capped entity store, only the current task authors' profiles, and the bounded 50-row recent-activity index. A task summary remains available by explicit ID through `read_task_summary`, and isolated telemetry directories no longer affect search cost.

src/connectors/slack/client.ts
hardworker: Fixed in `1d2552f` by reverting this refactor. `isChannelShared` again short-circuits `D…` conversations and keeps its existing advisory cache; it no longer claims to share a snapshot with `getChannelInfo`. Visibility continues to come from the fail-closed `getChannelInfo` path.

src/memory/activity.ts
hardworker: Fixed in `1d2552f`. Prompt assembly now reads and injects `recent-activity.md` verbatim via `readActivityMarkdown`, so hand-edited or legacy rows are not silently dropped. Structured parsing remains only where entries are actually needed for bounded search and maintenance.

docs/plans/README.md
hardworker: Fixed in `1d2552f`. The plan index now says ‘bounded injection, pull tools, confidentiality layer’; the eval harness remains explicitly split to #228.

hardworker:
Addressed the review in `1d2552f`; all inline threads now have specific replies.

The important design call is telemetry: we intentionally retain full selection and pull telemetry for both public and private tasks. It now lives under `memory/telemetry/tasks/`, separate from the agent-readable corpus. Prompt assembly and memory tools never enumerate that subtree, while operators can still snapshot or download the complete `memory/` bundle in one batch. The deployment docs mark the telemetry subtree as sensitive.

The other blocking paths are closed as well: trigger tasks derive visibility instead of defaulting public, private tasks cannot approve persistent triggers, legacy/unverified triggers pause, and unmarked pre-boundary memory stores are refused rather than migrated. The first deployment requires a snapshot-and-clear reset.

Verification on the pushed commit: typecheck, build, and 933/933 tests across 69 files.

hardworker:
Replaced by #278 as the top PR in the reorganized Memory v2 stack. The reviewed head commit is unchanged; #276 and #277 now carry the lower semantic layers, and #228 remains the eval-only child.

PR #172, opened 2026-07-02
memory: Phase 1.5 — roadmap rescope + spawn selection sensor

hardworker:
## Summary

- **Roadmap rescope** (`docs/proposals/memory-v2-roadmap.md`): Phase 1.5 becomes an enablement pre-flight + permanent sensors instead of a replay-based eval gate (prod baseline is injection-off — no incumbent recall to protect; replay can't see the recency-ossification loop). A new Phase 5 memory-value eval runs on accumulated live data; earlier phases plant the sensors it consumes.
- **Selection sensor** (OpenSpec change `memory-v2-selection-sensor`, all tasks complete): every injection-enabled spawn appends one self-contained JSONL record (context snapshot, selected pages with scores, dropped, zero-signal count, budgets, token estimate) to the task's telemetry file. Fail-safe by construction; no new flags; zero cost while injection stays off.
- **Per-task layout** (follow-up decision in the same change): all per-task memory artifacts live in `memory/tasks/<taskId>/` — `summary.md` (moved from flat `memory/summaries/`, one-time idempotent startup migration) and `telemetry.jsonl` (new). Placement follows the writer: everything the memory layer produces lives — and ejects — with `memory/` (`rm -rf` stays single-step), `-m` memory-only pulls carry telemetry, and the tree now encodes the episodic (per-task) vs semantic (`entities/`, `users/`) split. `docs/architecture/memory.md` gains a dedicated Telemetry section (where/how/why).

## Verification

- `npm run typecheck` clean; `npx vitest run src/memory/__tests__/` — 332/332 (new: telemetry fields, record shape, one-line-per-spawn, fail-safe warn path, disabled-injection no-write, missing-taskId skip, migration move/no-op/skip, dot-traversal guard).
- In-app (pre-relayout, record semantics unchanged): dockerized worktree instance with the 139-entity store seeded and injection on; task created and observed via the `archie-debug` MCP. The PM spawn wrote a correct zero-signal record — `candidates: 0`, `zeroSignalExcluded: 138`, `selected: []`, `renderedTokensEst: 8207` (the always-injected index) — while the PM answered an Adjust/antifraud question correctly from index `L0` summaries alone (the recall safety net working as designed).
- Real-module harness against the final layout: `initMemory()` migrated a seeded legacy summary (and removed the emptied legacy dir), and an enriched spawn wrote the v1 record to `memory/tasks/<taskId>/telemetry.jsonl` with the task dir auto-created; signal-bearing selection score 510 = repo 500 + title-token 10.

## Notes for review

- The destination moved during review discussion: session-dir colocation was implemented first, then superseded by `memory/tasks/<taskId>/` before anything writes records in prod — rationale in the change's `design.md` (Decision 2, including why `tasks/` rather than `sessions/` naming).
- Telemetry appends are the one documented write outside the serialized lifecycle queue — safe because append-only, per-task, no read-modify-write.
- A CLI-created task spawns the PM with no title → zero signals; title generation currently runs only on the Slack path. The sensor makes this visible; worth watching in prod records.

https://claude.ai/code/session_01FbC7H29sppbVu18xBAB8d8


hardworker:
Rebased directly onto `feature/memory-v2`: the 4 selection-sensor commits (roadmap rescope, sensor, telemetry docs, per-task layout) now live on `feature/memory-v2` under new SHAs (`5d8bbe3`, `54752b0`, `cfa34b0`, `b4eae24`), on top of current `main` (which carries the #176 trust-boundary fixes). History is linear — no merge commit. `feature/memory-v2` is not merged to `main`. Closing; commits are integrated.

Issue #168, opened 2026-07-01
Human-in-the-loop approval gate for critical MCP tools


egorkhmelev:
## Problem

Today Archie treats MCP tools as all-or-nothing. A tool is either allowed for an agent (and then invoked with no human check) or it is restricted via `disallowedTools`/omitted from the `tools` allowlist. Because of this binary, we tend to **blindly restrict any tool that mutates an external system — or never wire it up at all**. There is no middle ground where a genuinely useful but sensitive tool (send a Slack message, create/update a Monday item, publish an offer in the admin panel, update a Notion page, delete a file on GitHub, merge a PR) can be exposed *behind a human approval*.

We want a **human-in-the-loop approval gate for a curated subset of critical MCP tools**, conceptually similar to the existing edit-mode request, so these tools can be safely enabled instead of blocked.

## Current state (for reference)

- The SDK runs in `bypassPermissions`. Tool gating is entirely static per spawn: the `tools` allowlist, the `disallowedTools` denylist, the OS sandbox, and a `PreToolUse` filesystem-guard hook. See `src/agents/spawn.ts`, `src/agents/sandbox.ts`.
- **There is no `canUseTool` / dynamic per-call permission callback anywhere today.** This is the main gap this feature has to close.
- Edit mode (`docs/architecture/edit-mode.md`) is a *task-level, one-way boolean* (`metadata.edit_allowed`): PM calls `request_edit_mode` → Slack Approve/Deny buttons (`src/connectors/slack/events.ts`, with a CLI/API path in `src/connectors/api/routes.ts`) → `handleEditModeApproval()` in `src/tasks/task.ts` flips the whole task to RW at the next spawn. It approves a *capability for the task*, never a specific action, and cannot be revoked.
- MCP servers come from a single `PLUGINS_DIR/.mcp.json`; agents opt in via `mcpServers` frontmatter (`src/system/plugin-loader.ts`, `src/agents/registry.ts`). Archie already extends the spec cleanly by carrying a non-standard `description` per server and **stripping it before** handing config to the SDK — the precedent for how an Archie extension should not break the plugin spec.
- Generic (non-repo) plugin agents can write to their own workspace `cwd`, but they have **no path to write to external systems** — every outbound mutation goes through MCP tools, which are currently either blocked or never enabled.

### Approval granularity — per-call, with arguments shown
[...]

PR #173, opened 2026-07-03
Keep per-user MCP OAuth in Slack DMs

hardworker:
## Summary

- Limit per-user MCP OAuth to tasks whose default channel is a 1:1 Slack DM.
- Keep usable shared credentials as the DM default and escalate only the failing server to the DM participant.
- Make authorization PM-only, deduplicate pending requests, carry challenged scopes, and bound forced reauthorization.
- Persist successful callbacks as an encrypted wake outbox that survives teardown races and daemon restarts.
- Bind personal grants and shared DCR clients to issuer, resource, and redirect URI.
- Harden every outbound OAuth request against SSRF, unsafe redirects, oversized responses, and stalled endpoints.
- Add RFC 9728 path/root discovery fallback, safe targeted revoke parsing, and resilient Slack DM identity lookup.
- Sanitize dropped OAuth-binding errors into actionable messages without logging provider payloads or user identifiers.
- Add an opt-in Notion live gate for shared-to-personal escalation, callback wake-up, restart, refresh, and targeted revoke with redacted evidence.

## Why

The earlier implementation selected personal credentials too broadly and inferred task teardown from transient task metadata. It could lose or duplicate authorization continuations around teardown, reuse credentials after a server binding changed, and let discovered OAuth endpoints bypass network-safety checks. The revised flow uses the stable DM participant boundary, explicit task termination state, and a durable at-least-once wake record.

## Impact

Non-DM tasks and legacy shared-token injection remain unchanged. DM tasks continue with shared credentials while they work; after a 401, 403, or insufficient-scope failure, the PM can switch only that server to a bound personal grant. Callback recovery can replay one continuation nudge after a crash, but cannot lose the encrypted grant or wake intent.

## Validation

- Rebased onto `origin/main` at `b7f9b56`
- TypeScript `7.0.2`
- `npm run typecheck`
- `npm run build`
- `npm test -- --run --reporter=dot` — 1,208 tests passed; 5 skipped
- `npx openspec validate per-user-mcp-oauth --strict`
- `git diff --check`
- `npm run e2e:oauth:notion --` — preflight/usage path verified

Lint was intentionally excluded. The real-provider Notion gate is implemented but remains unchecked because completing it requires live Slack and two separate Notion identities.


egorkhmelev:
@hardworker Maybe we should only allow personal MCPs in DMs? Not sure about doing it more widely. Do you have example in mind?

egorkhmelev (changes requested):
Head `945e279`. 1 blocking defect, 11 confirmed non-blocking, 4 scope questions. QA: 24/39 ACs verified, 11 failed, 4 unverifiable in this environment.

## Blocking

**`wakeDmTask` arms the resume on `task:stopped`, but the normal park path is `complete()`** — `src/connectors/oauth/routes.ts:241`

DM task escalates → user replies in the DM before clicking (task reactivates) → user clicks → `completeFlow` stores the token, `wakeDmTask` takes the `task.isActive` branch, registers a `task:stopped` listener, returns `true`, page says "The task resumes automatically" → PM ends its turn → `report_completion` → idle-check → `Task.complete()` emits `task:completed`. Listener never matches: the task is never resumed with the token just stored, and since `offEvent` is only reached on a match the closure leaks for the process lifetime (bus cap 50, `src/system/event-bus.ts:29`). Same hole in the mid-`complete()` window. Reproduced: `wakeDmTask` → `true`, `task:completed` → 0 `sendMessage`; a later `task:stopped` for the same id still fires the listener. Wake on terminal events generally, not just `stop()`.

## Non-blocking

1. **Escalation never rolled back** — `src/tasks/task.ts:1540` is the only write to `metadata.mcp_personal_oauth` and is add-only; no delete in `src/`. Abandoned authorization strands the server: deleted from `mcpServers` on every later spawn, shared credentials never restored (`injected: []`, `sharedInjected: []`, `requestable: ['notion']`, stable across spawns).
2. **No idempotency guard** — `src/tasks/task.ts:1546`. Already-escalated + valid token → `'ready'` → stop and respawn with identical credentials. A 403 from an unshared resource (not a scope gap) loops to the wall-clock cap. `request_edit_mode` guards this shape at `src/agents/tools.ts:612`.
3. **Resume hardcodes `pm-agent`** — `src/agents/tools.ts:714`, `routes.ts:229`. A repo/plugin specialist that escalated is never re-engaged, contradicting "you will be re-activated with access afterwards"; PM gets a continuation for a server it may not hold.
4. **No `task.suspendStatus()`** — `src/agents/tools.ts:723`. Measured: `request_edit_mode` 1 call, `request_mcp_auth` 0. "is working…" resurfaces on a parked task. Siblings suspend at `tools.ts:668/806/863/1868`.
5. **`toError()` erases the cause** — `src/system/oauth/inject.ts:113` returns `new Error(err.name)`; no OAuth error class assigns `name`, so `message === 'Error'` and `spawn.ts:575` renders `OAuth bind failed: Error`. Regression from `ebc7809`. And `inject.ts:51/67/73` still pass raw `err` to `logger.error`, so the exposure this sanitizing was for remains.
6. **Unbounded `fetch` on the spawn path** — `src/system/oauth/discovery.ts:29`, awaited at `inject.ts:84`. No `AbortSignal`, no dispatcher timeout. Deployed `plugins/.mcp.json` has three credential-less servers probed serially; one blackholed host stalls every DM spawn ~300s. The 60s error TTL does not help — nothing caches until the probe settles, and there is no in-flight dedupe. Add `AbortSignal.timeout`.
7. **`hasExplicitAuthHeader` early `continue` stops token refresh** — `src/system/oauth/inject.ts:41`. Previously operator-header servers still ran `ensureFreshToken`; now the vault record ages out (no keep-warm), and the first spawn after the header is removed fails its refresh. Out of the PR's stated scope.
8. **`request_mcp_auth` on the shared base server** — every agent gets it (`tools.ts:2778`, `spawn.ts:294`) and `task.requestMcpAuth` calls `postToUser` (`task.ts:1607`), against `docs/architecture/agents.md:48`/`:57` PM-only messaging and PR #232's constraint. `tool-contract.test.ts:116` asserts it, so likely deliberate — worth ratifying explicitly.
9. **`pendingTeardown` early-return states an untruth** — `src/agents/tools.ts:689`. `deferTeardown` is first-wins, also armed by `request_edit_mode`/`request_max_mode`/`request_merge_approval`. Returns "Authorization request already sent" when no link was posted and no pending record exists.
10. **Docs** — `docs/architecture/secrets.md:124` promises an example with a trailing colon; the next line is the next heading. `secrets.md:233` says the pending record persists only `state`/`task_id`/`slack_user_id`; `connect.ts:311-326` also persists issuer, endpoints, scopes, resource, redirect_uri and created_at in plaintext.
11. **Dead code and churn** — `resetServerAuthClassCache()` (`discovery.ts:133`) has no caller. `resolveRedirectUri()` duplicates logic still inlined at `cli/oauth.ts:79-86`. `--ignore-blank-lines` drops `slack/events.ts` from the diff entirely (same for one line each in `types/task.ts` and `oauth/types.ts`). `client.ts:1286` returns `imUserId` and `dmUserId` holding the identical value.

## Before merge

- `openspec/changes/per-user-mcp-oauth/tasks.md:30` (`4.4 Verify the complete flow against a real OAuth MCP server`) is unchecked — the headline path browser → `/oauth/callback` → `writeUserOAuthRecord` → wake → inject has no coverage; `routes-wake.test.ts` mocks `Task.get`.
- Rebase and re-validate: main moved TypeScript `^6.0.3` → `^7.0.2` after the last commit here, and this PR touches neither `package.json` nor the lockfile, so its ~1,300 new TS lines have never been compiled under TS 7.
- CodeQL alert #24 is `js/clear-text-logging` at `src/system/logger.ts:281`, dismissed as a false positive — not in `inject.ts`, and it cannot transition to resolved.

## Questions

1. Proactive spawn-time probing of credential-less servers is a second escalation trigger beyond "agent hits 401/403" — intended, or should escalation stay reactive? It is the source of both the timeout risk and the cache machinery.
2. Narrow `request_mcp_auth` to the PM?
3. Bespoke DM escalation vs issue #168's general per-call approval framework — and is `channel_id.startsWith('D')` in core acceptable while #203 reworks the channel/identity abstractions?
4. Deleting main's two documented non-goals from `secrets.md`: the DM-only narrowing rests on one 2026-07-05 comment answered by a force-push with an empty body. Currently 0 reviews, 0 inline comments, no reply.

PR #191, opened 2026-07-07
fix(tasks): restart repo agents on edit-mode approval for a writable mount

egorkhmelev:
## Problem

A repo agent's read/write access **is** the bwrap sandbox, decided **at spawn time**: `editAllowed` puts the clone in `allowWritePaths` (RW) vs `denyWritePaths` (kernel-RO). [`handleEditModeApproval`](../blob/main/src/tasks/task.ts) sets `edit_allowed = true` and messages PM — but a repo agent that is **already running** keeps its read-only mount and never re-reads the flag. So every write hits a **read-only filesystem** after approval.

### Observed — `task-20260625-1122-30wkzk` (IO-3070 BE PR)

Edit mode approved 11:56 (`[system] Edit mode approved by Alex Negru`), yet 11:56 → 12:19:
- `create_branch` denied 3×
- `touch` in repo root → **`Read-only file system`** (EROFS)
- `git status` → EROFS on `.git/lfs/tmp`
- `EnterWorktree` registered a worktree but it landed on the same RO mount
- Current Context tagged the repo `[READ-ONLY]`

No PR was ever pushed; the fully-staged implementation was handed off as an artifact. The user's read: *"BE agent was never restarted after approve."* Exactly right.

## Fix (in `handleEditModeApproval`)

1. **Flush `edit_allowed` synchronously** — `await save(true)` instead of `debouncedSave()`. A re-spawn (here, or after a park/reload) reads the flag from disk, so it must be persisted before any spawn, not 500 ms later.
2. **Tear down any live repo agent** — `abort()` + `queue.stop()` + drop from `agentProcesses`. PM's delegation then re-spawns it **fresh** with a writable clone. It resumes the **same session** (`session_id` persists in metadata), just RW. PM is not a repo agent, so it is left running.

## Test

`edit-mode-restart.test.ts`: a running repo agent is torn down + removed on approval, PM kept, `edit_allowed` set; an idle repo agent is left alone. Mutation-checked — disabling the teardown loop makes it fail (`abort` not called, agent not removed). `spawnAgent` mocked, fs writes stubbed. Suite 627/627, typecheck clean.

## Verification

Unit test covers the teardown logic; the actual **writable re-mount** is best confirmed with the `archie-e2e` **edit-mode scenario** against a live instance.

## Note (separate, unresolved)

Final metadata for the incident task has `edit_allowed` **entirely absent** — set in memory at 11:56 (finding proves it) but never persisted, with **no** concurrency signature in the transcripts. `save(true)` here closes the persistence window regardless, but the exact loss mechanism isn't pinned from the session dump — worth a forensic pass with the host app logs.


egorkhmelev:
Folded in the follow-up fix from #198 as `bbbbe10` (clean fast-forward onto this branch). This PR now carries **both** cases:

1. the original: a repo agent **already running** at edit-mode approval is torn down + re-spawned writable (this PR's first commit);
2. new: a reconcile in `ensureAgentSpawned` for a repo agent still **booting** at approval — the mid-boot window this branch's `isRunning`-guarded loop misses, where the agent finishes booting read-only and every `create_branch`/git write is then denied.

CI is green on the new head. #198 (which held that commit) is closed in favor of this PR.

---

PR #198, opened 2026-07-08
fix(tasks): restart repo agents that boot read-only during edit-mode approval

egorkhmelev:
## What & why

Builds on #191. That PR fixes the case where a repo agent is **running** when edit mode is approved — it tears the agent down so PM's delegation re-spawns it with a writable mount. But its teardown loop guards on `agent.isRunning`, which is `false` while an agent is still **booting** (the SDK handle isn't assigned until `spawnAgent` finishes, after the multi-second clone setup).

So a repo agent that is **mid-boot at the instant of approval** slips past the loop, finishes booting read-only, and — because the sandbox mount *and* the repo-tool allowlist (`create_branch`, `push_branch`, …) are frozen from `edit_allowed` at spawn time — can never write. Every git write returns `Permission to use mcp__repo-tools__create_branch has been denied` and the task stalls. This is exactly the "PM asks a follow-up that spawns the engineer, then the user approves while it's working" race that motivated #191.

Reproduced against a live instance: approving at the moment the agent began cloning left it read-only, `create_branch` denied twice, and the requested change was never made.

## How it works

`Task.ensureAgentSpawned` reconciles the stale agent the next time work is delivered to it (which is the only moment a read-only agent can actually do harm — its next turn). If the existing agent is a repo agent that booted read-only (`editModeAtSpawn === false`) while `edit_allowed === true`, it is torn down (`abort()` + `queue.stop()`) and dropped, its session synced into metadata first, so the fresh spawn below comes up writable and resumes the same SDK session.

- New `Agent.editModeAtSpawn` records the `edit_allowed` value a process was built under; `spawnAgent` sets it in the repo branch alongside the sandbox/allowlist it gates.
- The reconcile deliberately reuses the established **abort + `queue.stop()` + drop + fresh-`Agent`** teardown shape (the same one `handleEditModeApproval` uses) rather than an in-place re-spawn. An earlier in-place attempt was rejected in review: aborting without stopping the queue trips the aborted boot's session-recovery on a still-live shared queue (wiping `session_id`, re-queueing messages), and re-spawning inside `Agent.spawn` races `handleEditModeApproval`'s own delete loop.

Out of scope: #191's running-agent path is unchanged; this only adds the mid-boot backstop. The two are complementary — "already running at approval" (#191) and "still booting at approval" (this) now both end writable.

## Verification

- `npm run typecheck` clean; full suite **630/630** (adds 3 reconcile tests). The regression test is mutation-checked — disabling the reconcile fails it — and the negatives pin the already-writable and non-repo cases.
- **Live**, against an instance booted from this branch (archie-e2e harness): drove the exact race — PM parks at the edit-mode gate → a follow-up makes it delegate a read-only investigation → approval fired at clone-start (mid-boot). The reconcile fired (`Edit mode approved — restarting archie-agent for a writable mount`), there were **zero** permission-denied errors, and the re-spawned `:rw` agent ran `create_branch` → edited README → committed (`0fadb09`) → task completed. Pre-fix, the identical timing produced repeated denials and a stall.
- Honest gap: the `git push` in the live run failed because the sandbox has no GitHub-App push credentials — unrelated to this change (the writable mount + local commit are what this fixes). Session-resume continuity is preserved in code and asserted in the test; in one live run the re-spawn cold-started because the read-only turn's `session_id` wasn't yet on the agent object at reconcile time — functionally harmless (PM re-sends the instruction), noted for follow-up.

---
_Generated by [Claude Code](https://claude.ai/code/session_01YKSaELJsA2PDodV11vpxKc)_


egorkhmelev:
Folded into #191. Its branch `fix/edit-mode-approval-restart` now contains this commit (`bbbbe10`) via fast-forward, so #191 carries both the running-agent restart and this mid-boot reconcile. Closing in favor of #191 — no code lost.

---

PR #206, opened 2026-07-10
Add triggers — persistent "do Y when X happens" rules

egorkhmelev:
## What & why

Archie was purely reactive — it only acted on an inbound Slack message or a GitHub webhook. The one bit of proactivity (the reminder scheduler) could re-wake an existing task but never start work on its own. This PR adds **triggers**: persistent, user-approved "do Y when X happens" rules that spawn a fresh task when a condition fires. It's the one sanctioned form of self-initiated work, and every trigger is created behind an explicit Approve/Deny gate.

Two condition types ship in v1: **schedule** (a recurring cron cadence or a one-off future time) and **channel-message** (a new top-level message in a watched channel, with an optional substring / author filter). Each trigger is bound to a delivery target — a channel or a user DM. GitHub-event triggers are a deliberate follow-up.

## How it works

The design reuses existing machinery rather than adding a new engine. The firing loop is a near-clone of the reminder scheduler's index-and-tick pattern (`trigger-scheduler.ts`): an in-memory index of enabled triggers, a 60s tick, and an immediate catch-up check on boot. Recurrence math is offloaded to `croner` (DST-correct, zero-dep) and kept **entirely internal** — the PM translates natural language → a cron expression at creation, the system validates a ≥1h floor for recurring schedules, and listings render cron back to prose. One-off schedules store a single `next_run_at` and auto-pause after firing. Triggers persist one JSON file each under `$ARCHIE_WORKDIR/triggers/` (`trigger-store.ts`).

Creation reuses the edit-mode approval mechanism, which is already channel-agnostic: `propose_trigger` posts an Approve/Deny prompt that renders as buttons in Slack and as `[y]/[n]` in the CLI, and both converge on the same task-level handlers (`handleTriggerApproval`/`handleTriggerDenial`) — the Slack button and the CLI `/tasks/:id/approve` endpoint. There is no operator bypass. The proposed trigger id is threaded through the approval event so concurrent proposals resolve to the right one.

Fired tasks are ordinary read-only tasks: `fireTrigger` spawns a task, wires the delivery destination, sets `triggered_by`, and seeds the PM with the action prompt. Firing posts no preamble — the spawned PM does the work and posts the result itself. A provenance gate blocks a trigger-spawned task from creating more triggers (no amplification loops). Visibility is scoped by the tier of the space the request comes from (public / private / DM), with channel privacy resolved from the workspace channel cache (`listWorkspaceChannels`) plus a live fail-closed fallback on a miss, so a public→private conversion is reflected promptly and a private trigger is never leaked into a public listing. Config changes (create/edit/pause/resume/delete) are announced to the bound channel; firing is not.

The PM gets four tools (`propose_trigger` / `list_triggers` / `update_trigger` / `delete_trigger`), an engine-owned skill (`skills/triggers`) plus a short always-present blurb in the PM base prompt so it knows triggers exist before loading the skill, and the operator surface gets `/api/triggers` endpoints and a CLI trigger-list view (press `t`). Protections: the ≥1h recurring floor, per-user and per-channel active-trigger caps (re-checked on resume), a per-account daily fire cap, a dead/archived-channel pre-flight that pauses and notifies, stale-pending GC on boot, and an `ARCHIE_TRIGGERS_ENABLED` kill switch.

Out of scope on purpose: GitHub-event triggers, and any user-facing cron surface.

## Verification

Static checks all pass on the rebased tip: `npm run typecheck`, `npm run build`, and `npm test` (**782 tests, 59 files**), including new unit tests for cron next-run (with a DST boundary), the ≥1h-floor validator, and the visibility-scoping decision.

I did **not** boot the live app / run the E2E harness in this environment, so the end-to-end Slack and CLI paths (button approval, `[y]/[n]` approval, a real fire posting into a channel/DM) have not been exercised against a running instance — worth a live check before relying on it.

One thing to look at closely in review: this branch was written before `main`'s `postToUser` refactor and rebased onto it. That refactor narrowed `PostTarget` to `channel` only (removed `new_dm`/`new_thread`) and deleted `openDMChannel`, routing cross-channel posting through the new task-decoupled `post_to_channel` explore tool. I adapted trigger delivery to the new model — system-level announcements/notices post directly via `postSlackMessage` (a Slack user id resolves straight to that user's DM, so no `conversations.open` is needed), and `fireTrigger`'s delivery instruction to the PM is now tool-agnostic plain language ("post the result to #channel" / "as a DM to the user"). The PM-driven delivery of a fired result therefore leans on whatever posting tools the PM currently has under the new model; that's the specific path I'd validate live.

## Deployment & follow-ups

- No migration. Triggers live in a new `$ARCHIE_WORKDIR/triggers/` directory created on boot.
- New dependency: `croner`.
- `ARCHIE_TRIGGERS_ENABLED=false` disables all firing and creation globally (kill switch); default is enabled.
- Follow-up: GitHub-event triggers (webhook plumbing already exists); and a live E2E pass over the approval + fire paths.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MzcVLThyuwSpo8p6RTE2wT)_

### DIFF
--- a/src/agents/tools.ts
+++ b/src/agents/tools.ts
@@ -34,6 +34,10 @@ import {
   fetchExploreThread,
   postSlackMessage,
   assertPostableChannel,
+  getChannelInfo,
+  getUserInfo,
+  listWorkspaceChannels,
+  fetchChannelIsPrivate,
 } from '../connectors/slack/client.js';
 import { readCanvas } from '../connectors/slack/canvas-read.js';
 import { collectCanvasFileAllowlist } from '../connectors/slack/channel-canvas.js';
@@ -87,6 +91,32 @@ import { scheduleReminder, cancelReminder } from '../system/reminder-scheduler.j
 import * as chrono from 'chrono-node';
 import { writeFile } from 'fs/promises';
 import { join } from 'path';
+import type { Trigger, TriggerBinding, TriggerCondition } from '../types/trigger.js';
+import {
+  generateTriggerId,
+  saveTrigger,
+  loadTrigger,
+  listTriggers,
+  deleteTrigger,
+  countActiveTriggers,
+} from '../system/trigger-store.js';
+import {
+  computeNextRun,
+  validateRecurringInterval,
+  planStatusChange,
+  indexTrigger,
+  deindexTrigger,
+  announceTriggerChange,
+  describeTrigger,
+  triggerWhat,
+  triggerWhen,
+  triggerWhere,
+  triggersEnabled,
+  MAX_TRIGGERS_PER_USER,
+  MAX_TRIGGERS_PER_CHANNEL,
+} from '../system/trigger-scheduler.js';
+import { emitEvent } from '../system/event-bus.js';
+import { triggerVisibleFrom, type TriggerOrigin } from '../system/trigger-visibility.js';
 
 // Re-export branch state helpers for consumers that import from tools.ts
 export { hydrateBranchState, findBranchStateByPR };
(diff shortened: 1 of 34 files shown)


egorkhmelev:
Addressed the 5 CodeQL "uncontrolled data in path expression" alerts in `src/system/trigger-store.ts` (all the same root cause: a filesystem path built from a caller-supplied trigger id — the PM tool arg, the `/triggers/:id` route param, an approval `ref`, and the Slack button value all flow in).

Fix: added a strict id validator (`isValidTriggerId`, `^trg-[A-Za-z0-9-]+$`) — the charset excludes `.` and path separators, so a validated id can't express traversal. `getTriggerPath` now throws on a malformed id (the single choke point), and `loadTrigger`/`deleteTrigger` validate up front and treat a malformed id as "not found" / no-op. Added unit tests covering the traversal cases. Pushed as `c3e5c91`; CodeQL should clear on the re-scan.

---

src/system/trigger-scheduler.ts
```diff
@@ -0,0 +1,375 @@
...
+  task.metadata.triggered_by = trigger.id;
+
+  // Wire delivery. message context → reply in the triggering thread (linked as
+  // default, no post). schedule context → the PM opens the destination itself.
+  let delivery: string;
+  if (context.kind === 'message' && context.channelId && context.threadId) {
+    task.linkSlackThread(context.channelId, context.threadId, context.channelName ?? context.channelId);
+    delivery = 'Post your reply in your default channel — the thread where the triggering message was posted.';
+  } else if (trigger.binding.type === 'user') {
+    delivery = `Deliver the result to the user as a direct message (Slack user ID ${trigger.binding.user_id}).`;
```
egorkhmelev, line 373: **H1 (High) — a schedule trigger bound to a user DM never delivers its result.**

A schedule-context fire with a `user` binding wires no channel onto the spawned task (only the message-context branch calls `linkSlackThread`). The fired PM's only DM tool is `post_to_user`, which drops the message when the task has no linked/`default_channel` (see `postToUser` in `src/tasks/task.ts` — the "no default channel — message dropped" path). `post_to_channel` explicitly rejects user/DM ids. So there is no tool path for the fired PM to DM a user.

Confirmed live (booted from this branch): a user-bound one-off fired → a read-only PM task spawned with `triggered_by` set → the PM called `post_to_user message=TICK-…` with the correct content → the task completed and **nothing was delivered**. The drop is structural — `postToUser` returns before any Slack call — so it fails identically in production, not just in a Slack-less test instance. Channel-bound and message-triggered delivery both work.

Fix direction: expose a DM-capable delivery to the fired PM (the internal announce path already DMs via `postSlackMessage({ channel: user_id })`), or reject `user` bindings at `propose_trigger` until this is supported.

---

src/agents/tools.ts
```diff
@@ -2010,6 +2035,312 @@ function createFetchSlackReferenceTool(agent: Agent, task: Task) {
...
+  };
+  return (channelId: string) => {
+    let p = cache.get(channelId);
+    if (!p) {
+      p = workspaceMap().then((map) => {
+        if (map.has(channelId)) return map.get(channelId)!;
+        // Miss — a brand-new, just-converted, or archived channel not in the
+        // cached list. Resolve it live and fail closed (treat as private on
+        // error) so a private channel is never leaked into a public listing.
+        return getChannelInfo(channelId).then((i) => i.isPrivate).catch(() => true);
```
egorkhmelev: **H2 (High) — the privacy resolver fails *open*, not closed.**

This `.catch(() => true)` is meant to fail closed (treat an unresolvable channel as private), but `getChannelInfo` swallows every `conversations.info` error internally and returns `{ isPrivate: false }` (the catch block in `src/connectors/slack/client.ts`). It never rejects, so this `.catch(() => true)` is dead code and the miss path resolves to **public**.

Sequence that leaks: a trigger is bound to a private channel → the bot is later removed from it → the channel drops out of the `listWorkspaceChannels` cache (map miss) → the live `conversations.info` lookup errors → the resolver returns "public" → `list_triggers` run from any public channel or DM now exposes that private trigger's id, status, full action-prompt text, and `#channel_name`. This contradicts the fail-closed guarantee stated in AC7 and `docs/architecture/triggers.md`, and turns the documented ≤10-min staleness into a permanent leak.

Fix: add a `getChannelInfo` variant that rethrows on error (so this `.catch` runs), or have the miss branch treat any resolution error as private.

---

src/system/trigger-scheduler.ts
```diff
@@ -0,0 +1,375 @@
...
+    enabledTriggers.delete(trigger.id);
+    return;
+  }
+  const now = Date.now();
+  for (const cond of trigger.conditions) {
+    if (cond.type !== 'schedule' || !cond.cron) continue;
+    const due = cond.next_run_at ? new Date(cond.next_run_at).getTime() : 0;
+    if (!cond.next_run_at || due <= now) {
+      const next = computeNextRun(cond.cron, cond.tz);
+      if (next) cond.next_run_at = next.toISOString();
```
egorkhmelev: **M1 (Medium) — overdue recurring triggers never fire on boot, contradicting the catch-up promise.**

On boot, `rebuildFromDisk` → `indexTrigger` advances a recurring condition's past `next_run_at` to the next *future* occurrence here, and this runs *before* the immediate `checkDue()` in `initTriggerScheduler`. So `checkDue` sees a future `next_run_at` and skips it — the run missed during downtime never fires. Only one-off conditions catch up (they're exempt from this advance because the guard requires `cond.cron`).

But `initTriggerScheduler`'s docstring and `docs/plans/v31-triggers.md` both describe an immediate check that "fires overdue schedules after downtime." Example: a daily 09:00 digest with the process restarting at 09:05 silently skips that day.

Either fire once on catch-up (detect that the previous window was missed before advancing), or correct the docstring/plan to state that recurring schedules do not fire retroactively. The contradiction between documented and actual behavior is the defect.

---

src/system/trigger-scheduler.ts
```diff
@@ -0,0 +1,375 @@
...
+          await notifyCreator(trigger, `⚠️ A trigger you set up was paused — its schedule could no longer be computed. Recreate it if you still need it.`);
+          break;
+        }
+      } else {
+        // One-off: auto-pause after firing once.
+        trigger.status = 'paused';
+        await saveTrigger(trigger);
+        deindexTrigger(trigger.id);
+        emitEvent('trigger:paused', trigger.id, { reason: 'one-off fired' });
+        break;
```
egorkhmelev: **M2 (Medium) — a trigger mixing a one-off and a recurring condition loses the recurring forever.**

When a one-off condition fires, this branch sets the whole `trigger.status = 'paused'`, deindexes it, and `break`s out of the conditions loop. Any sibling conditions on the same trigger are never processed again. So a trigger like "remind me once tomorrow, **and** every Monday after that" permanently loses its weekly schedule the moment the one-off fires.

Related, a few lines up: `checkDue` fires once *per due schedule condition*, so two conditions due in the same tick spawn two tasks — against the "N conditions — any match fires the trigger" model in `types/trigger.ts`. (The ≥1h floor is per-condition, so two coincident `0 9 * * *` conditions aren't caught by it.)

Consider auto-pausing/advancing per-condition rather than pausing the entire trigger, and de-duping fires within a tick.

---

egorkhmelev (commented):
## 🤖 Archie / Forge review — recommendation: **Request changes**

Independent review run via Forge `review` mode: intent + ACs derived from the PR (no interview), two fresh-context reviewers (spec-compliance + adversarial bug-hunt), a QA coverage+suite pass, and a **live boot-from-branch E2E** (Docker instance built from `0f5b03b`, driven through the HTTP API). Every finding was verified in code. The author keeps ownership — this is advisory.

The design is sound and largely faithfully implemented. Blocking on two high-severity gaps and two medium correctness issues, plus a firing-engine test gap. Details are in the four inline comments (H1, H2, M1, M2); everything else is summarized below.

### Live E2E (booted from this branch, `/health` attested to `0f5b03b`)
- `basic-nonce`: **pass** — task completed, PM replied, nonce in log; the branch runs end-to-end through the proxy.
- Trigger lifecycle: **pass mechanics** — pending trigger correctly *absent* from `GET /api/triggers` before approval (gate holds); the HTTP `/approve` `type=trigger` path flipped it to `enabled`; a one-off **fired and auto-paused** (`last_fired_at` set); a **read-only** PM task spawned with `triggered_by` set.
- That same run surfaced **H1** (see inline): the fired user-bound task called `post_to_user` with the right content and delivered nothing.

### Inline findings
- **H1 (High)** — `user`-bound schedule triggers never deliver their result (no DM tool path for the fired PM). Confirmed live.
- **H2 (High)** — privacy resolver fails *open*; a private-channel trigger can leak into public/DM `list_triggers` after the bot is removed from the channel.
- **M1 (Medium)** — overdue recurring triggers don't fire on boot, contradicting the "catch-up after downtime" docstring/plan.
- **M2 (Medium)** — a trigger mixing a one-off with a recurring condition permanently loses the recurring when the one-off fires.

### Low severity
- **L1** — caps aren't re-checked at approval: proposing N triggers while under the cap (pending ones don't count), then approving all, exceeds the per-user/channel cap of 20 (it's only re-checked on resume).
- **L2** — a fire bailed by the dead/archived-channel pre-flight still consumes a daily-cap slot (`withinDailyCap()` increments before the pre-flight bail); a multi-condition unreachable-channel trigger also DMs the creator once per due condition.
- **L3** — an unhandled throw in `checkDue`'s post-fire block (e.g. a `saveTrigger` disk error) unwinds the loop and skips the remaining triggers that tick (self-heals next tick).
- **L4** — `indexTrigger` mutates `next_run_at` in memory without persisting; benign today but a latent divergence with the PATCH/`update_trigger` save-then-index order.
- **L5** — `setInterval` doesn't await the async tick; a tick running >60s could let the next tick double-fire a not-yet-advanced condition (plausible, not confirmed).

### Test coverage
Suite is green — **788 tests / 60 files** (the PR body's "782 / 59" is stale; all pass). Pure helpers (cron next-run incl. DST, the ≥1h floor, id validation/traversal, the visibility matrix) are well covered. But the **entire firing engine is untested**: `checkDue`/`fireTrigger`/`indexTrigger`/`rebuildFromDisk`, one-off auto-pause, the daily cap, the dead-channel pre-flight, boot catch-up, loop protection, and the kill switch. M1 and M2 — and a regression that re-enabled trigger-spawned triggers — would all pass CI today.

### Verified good
Approval-gate integrity (no operator bypass; Slack button and CLI/API `/approve` converge on the same task handler); concurrent-proposal resolution (`ref` threaded end-to-end on both paths); the **path-traversal fix is watertight** (strict id regex + resolved-path containment + all store entry points validate); kill switch honored at creation, firing, and tick; fired tasks are read-only; timezone/DST handling; and the ≥1h floor sampling (`nextRuns(6)`) is sound.

### CI / merge
The CodeQL check is red and the PR is `mergeable_state: blocked`, but the path-traversal remediation is correct and complete in code and the alert threads are resolved with the Analyze job passing — this looks like a re-scan / alert-dismissal lag rather than a live vulnerability. Worth confirming the head re-scan clears the 5 alerts before merge.

### Assumptions (correct me if wrong)
1. I treated **user-DM delivery as a hard requirement** (the description lists it as a supported target). If it was meant as best-effort/future, H1 downgrades.
2. I treated **recurring-should-catch-up-on-boot** as the intent (per the docstring/plan). If "don't retroactively fire" is intended, M1 becomes a docs-only fix.
3. The live E2E ran with **no Slack configured** (CLI/HTTP entry), so real Slack channel/DM delivery and `channel_message` firing weren't exercised live; H1 was proven via the structural drop + container logs.

_Generated by Claude Code (Forge review mode)._

---

egorkhmelev:
Thanks — all findings addressed in `552c1ae`. Dispositions:

**H1 (user-DM never delivers)** — per the author decision, v1 is **channel-only**: `propose_trigger` no longer accepts `user` bindings, and the skill + `docs/architecture/triggers.md` now state DM delivery is a follow-up. (Creator notifications still DM fine — that's the system-level `postSlackMessage({channel: user_id})` path, not the fired PM.)

**H2 (privacy resolver fails open)** — fixed. Added `fetchChannelIsPrivate` in the Slack client that **rethrows** on error (unlike `getChannelInfo`, which swallows to `isPrivate:false`); the resolver's miss branch now uses it, so an unresolvable channel resolves **private** and the `.catch(() => true)` actually runs. A private trigger can no longer leak after the bot is removed.

**M1 (no boot catch-up)** — fixed to fire once, matching the docstring/plan. `indexTrigger` no longer advances a past `next_run_at`; the boot `checkDue` sees it as due, fires once, then advances past all missed windows via `computeNextRun(now)`.

**M2 (mixed condition loses recurring; double-fire)** — fixed. `checkDue` now goes through a pure `planTick`: it fires **once per trigger per tick** (dedupe) and expires one-offs **per-condition**, so a recurring sibling survives. Only when no schedule/`channel_message` condition remains does the trigger pause.

**Lows** — L1: caps re-checked at approval (pending no longer slip past). L2: daily-cap check moved after the reachability pre-flight (bailed fire doesn't burn a slot; one-fire-per-tick also removes the per-condition creator DM). L3: per-trigger try/catch in the tick. L5: re-entrancy guard so a slow tick can't overlap the next. L4 (index mutates without persisting) left as-is — benign: the computed value persists on the first fire.

**Tests** — added `planTick` unit tests covering the M1 catch-up and M2 per-condition/dedupe rules (suite now 794). The pure planning logic is covered; `fireTrigger`/`checkDue` end-to-end still lean on the live E2E rather than unit mocks.

**CodeQL** — the remaining single alert (down from 5) has resisted every canonical remedy (strict id regex, `basename`, resolved-path containment, regex-match value); it matches your read that the traversal fix is complete in code and this is a scanner false-positive/lag. Happy for it to be dismissed, or point me at the exact flagged source→sink if you'd rather I keep at it.

---

src/connectors/slack/client.ts
```diff
@@ -1393,6 +1417,26 @@ export async function fetchSlackFileBody(fileUrl: string): Promise<string> {
...
+ * Probe whether a channel can currently receive a post from the bot. Returns
+ * false when the channel no longer exists or is archived (e.g. deleted, or the
+ * bot was removed and the channel archived). A successful `conversations.info`
+ * on a live, non-archived channel returns true. Note: a public channel returns
+ * true even if the bot isn't a member (Slack allows posting), so this primarily
+ * catches the deleted/archived cases — the strongest signal available without
+ * actually posting.
+ */
+export async function isChannelReachable(channelId: string): Promise<boolean> {
+  const client = getSlackClient();
```
egorkhmelev: **N1 (Low, robustness) — `getSlackClient()` sits outside the try/catch, so `isChannelReachable` throws instead of returning `false` when the client is unavailable.**

The try/catch below is meant to make this a safe "reachable?" probe, but `getSlackClient()` on this line runs before it. When the Slack client isn't initialized, `getSlackClient()` throws — and `fireTrigger`'s channel pre-flight (`if (!(await isChannelReachable(...)))`) doesn't treat that throw as "unreachable → pause"; it propagates, the per-trigger `catch` in `checkDue` logs `Failed to tick trigger …`, and the due trigger neither fires nor pauses — it retries every 60s tick indefinitely.

Surfaced live: in a Slack-less E2E instance, an approved channel-bound one-off error-looped every tick (`Slack client not initialized` at `fireTrigger` → `tickTrigger` → `checkDue`) rather than pausing. Production impact is low — Slack is initialized at boot — but a transient client failure, or any deployment without Slack, turns every channel-bound schedule trigger into a per-tick error loop instead of a graceful pause/skip.

Fix: move `const client = getSlackClient();` inside the try (so a throw returns `false`), or have `fireTrigger`'s pre-flight treat a thrown reachability check as unreachable.

---

egorkhmelev (commented):
## 🤖 Archie / Forge review — round 2 (verified against `6f2e255`)

Re-verified every round-1 finding against the current head, re-ran the suite, and ran a second live boot-from-branch cycle. All the fixes hold.

### Round-1 findings — all resolved on `6f2e255`
| Finding | Status | Verified |
|---|---|---|
| **H1** user-DM delivery drops | ✅ Fixed | `propose_trigger` binding is `enum(['channel'])`; **confirmed live** — asked the PM for a DM-delivered daily summary and it refused (*"direct-message delivery isn't supported… I haven't created any user/DM-bound trigger"*) and created nothing. |
| **H2** privacy fail-open | ✅ Fixed | resolver miss-branch uses `fetchChannelIsPrivate` (rethrows) → unresolvable channel resolves **private**. |
| **M1** no boot catch-up | ✅ Fixed | `indexTrigger` leaves a past `next_run_at`; boot `checkDue` fires once. Unit-tested. |
| **M2** mixed one-off/recurring loss + double-fire | ✅ Fixed | pure `planTick`: one fire/tick, per-condition expiry, recurring sibling survives. Unit-tested. |
| **L1–L3, L5** | ✅ Fixed | caps re-checked at approval; daily-cap after pre-flight; per-trigger try/catch (**confirmed live** — isolated a fire exception, scheduler kept ticking); re-entrancy guard. |

### New in `6f2e255` — loop-protection redesign (worth a conscious sign-off)
The provenance gate was **removed**: a trigger-spawned task may now call `propose_trigger`. The rationale (docs + skill) is that human Approve/Deny is the real loop guard — a fired task can only create a `pending` trigger and can't self-approve, so there's no autonomous amplification. That reasoning is sound. One residual: a fired task can now emit *pending-proposal* approval prompts, and those are bounded only by the 24h stale-pending GC, not by the active-trigger caps (which count enabled) or the daily *fire* cap (which counts fires). Low risk — read-only, human-gated — but it's a deliberate posture change from "hard gate" to "approval-is-the-guard"; flagging so it's an explicit decision rather than a silent drop.

### New finding
- **N1 (Low, robustness)** — `isChannelReachable` calls `getSlackClient()` outside its try/catch, so it throws (rather than returning `false`) when the client is unavailable, and `fireTrigger`'s pre-flight lets that throw turn a due channel-bound trigger into a per-60s-tick error loop instead of a pause. Surfaced by the live run; low production impact (Slack inits at boot). See the inline comment on `client.ts` for the one-line fix.

### Live cycle (booted from `6f2e255`, `/health` attested)
- ✅ Branch boots and runs (`basic-nonce` completed, nonce in log).
- ✅ Channel-bound propose → (pending absent from `/api/triggers`) → approve via HTTP `/approve type=trigger` → `enabled`.
- ✅ H1 refusal (above) and ✅ L3 isolation (above).
- ⚠️ **Not verifiable here:** a channel-bound trigger actually **firing and delivering into a channel** — the headless harness has no Slack, and N1 makes the reachability pre-flight throw, so a real channel fire→deliver can't be observed. That path is covered only by unit tests (`planTick`) + static review. **Before merge, exercise one real channel-bound schedule fire against a Slack-connected instance** — no headless run can cover it.

### Suite & CI
- Local re-run on `6f2e255`: typecheck ✅, build ✅, **test ✅ — 806 tests / 61 files**.
- CodeQL still red (1 residual alert) and it's the only thing keeping the PR `blocked`. The traversal fix is textbook-complete in code (allowlist-regex-matched value + resolved-path containment + all store entry points validated); this reads as a scanner false-positive/lag — **safe to dismiss with a justification**.

**Bottom line:** every issue I raised is addressed; the code is green except the (spurious) CodeQL gate. Remaining asks before merge: (1) one live channel-fire smoke test on a Slack-connected instance, (2) decide on N1, (3) dismiss/clear the CodeQL alert.

_Generated by Claude Code (Forge review mode)._

---

Issue #203, opened 2026-07-10
Rework Slack into a pure UI renderer: engine emits semantic events, connectors subscribe


egorkhmelev:
## Problem

Slack is currently woven into the engine rather than sitting behind it. `tasks/task.ts` stores `SlackChannel`/`SlackThread` in task metadata, resolves channels, pushes Slack status, manages ack reactions, and builds message footers; PR cards live half in the Task class (`collectPrCards`, `refreshPrCardInPlace`, `resurfacePrCards`) and half in the Slack client's Block Kit builders. Adding any second surface (Discord, web, whatever) today would mean re-threading all of it — and even with no second surface planned, the coupling makes the engine untestable without a Slack workspace and lets renderer concerns (cards, status animation, titles) accrete inside task lifecycle code.

## Direction

Treat conversation surfaces as **subscribers**. The abstraction already half-exists: the engine emits semantic events on the event bus (`message`, `status`, `pr_card` with post/update actions, `approval:requested`, `task:*`), and the CLI/API is already a working second UI consuming exactly that stream. The rework is to make Slack consume the same vocabulary it currently bypasses:

- **Engine side:** emits semantic events out; accepts `sendMessage(taskId, text, agent?)` and approval resolutions in. Stores only connector-neutral state. Features like PR cards, live status, and titles become *events describing engine state*; how (and whether) they render is the subscriber's business.
- **Connector side:** `connectors/slack/` owns everything Slack — Block Kit, canvases, reactions, thread identity, update-in-place mechanics — as a renderer of the event stream plus an ingress that turns Slack events into `sendMessage`/approval calls.

**Approvals are semantic, not Slack-specific.** The CLI already renders `approval:requested` interactively and resolves it correctly — proof that the approval concept (a typed request with options and a resolution callback) belongs in the shared vocabulary as first-class. Each renderer maps it natively: Slack → buttons, CLI → interactive prompt. No lowest-common-denominator text fallback needed in the core contract.

## How to get there (no big bang)

1. **Dependency rule first.** Nothing outside `connectors/slack/` may import Slack types or the Slack client. Enforce with a cheap architecture test (grep-level: fail CI on `slack` imports outside the connector + `types/`). Land the test with an allowlist of current violations; shrink the allowlist opportunistically whenever a violating file is touched — no dedicated migration project.
[...]

Issue #202, opened 2026-07-10
Replace read-only GitHub tool wrappers with `gh` + a scoped read-only token; keep writes as policy-gated tools


egorkhmelev:
## Problem

`src/agents/tools.ts` (2,132 lines) hand-wraps ~25 GitHub operations as MCP tools. Most of them (`get_pr_reviews`, `get_pr_comments`, `get_review_threads`, `get_check_run`, `get_pr_checks`, `list_prs`, `list_branches`, `get_code_scanning_alert`, …) are read-only wrappers that re-implement `gh` one endpoint at a time. Every new read an agent needs means another wrapper, another schema, more engine surface.

The wrappers were never what enforced safety. Today's control is: (a) an application-level check — `resolveGithub` (`tools.ts:75`) rejects repos outside the agent's declared list — and (b) the fact that agents hold **no GitHub credential at all**: the sandbox env is explicitly minimal (`spawn.ts:604`, the SDK *replaces* env), and the single full-permission installation token only ever exists engine-side. Reads are harmless; wrapping each one buys no control, only code.

## Proposal

Enforce control at the **credential boundary** instead of the tool boundary:

**Reads — `gh` in the sandbox with a per-agent, read-only, repo-scoped token.** The engine mints an installation access token with `repositories: [the agent's declared repos]` and `permissions: { contents: 'read', pull_requests: 'read', checks: 'read', metadata: 'read' }` and exposes it to the sandbox as `GH_TOKEN`. The agent can run any `gh pr view / gh pr checks / gh api` incantation for investigation and *physically cannot write or reach undeclared repos* — the API refuses regardless of what the model types. Delete the read-only wrappers from `tools.ts`.

**Writes — unchanged, stay as gated MCP tools.** `push_branch` (branch-naming policy), `create_pull_request`, `merge_pull_request` (approval/auto-merge policy), `request_edit_mode`, comment/review-reply tools (attribution + logging). The write-capable token never enters the sandbox.

## Investigation: token generation (answering the on-the-fly vs pre-generated question)

[...]

Issue #201, opened 2026-07-10
Enforce a single canonical Task instance per taskId (task actor model)


egorkhmelev:
## Problem

Webhooks, Slack replies, the API/CLI surface, the merge orchestrator, and startup recovery can each construct their own `Task` instance for the same taskId. Every instance carries its own in-memory view of `metadata`, agent handles, and timers, so any two of them touching the same task race each other through disk.

This single design fact is the root cause of a whole family of correctness scaffolding that has accreted around it:

- `activationLock` in `src/tasks/task.ts` — keyed lock so concurrent triggers (GitHub webhook + Slack reply + recovery in one tick) don't double-spawn agents. Added after duplicate PR actions and looping agents were observed in production (2026-06-30).
- `cardLock` in `src/tasks/task.ts` — guards concurrent PR-card writes from separate Task instances.
- The "clear-before-awaits" invariant in `handleMergeApproval` (`task.ts:~1329`) — approval markers must be cleared before any `await` so a concurrently constructed instance doesn't observe stale state.
- Synchronous `save(true)` flushes (e.g. arming a PR for auto-merge, edit-mode `edit_allowed` before respawn) — needed because another instance may re-read metadata from disk at any moment.
- The merge orchestrator building "single Task instance per check, marker flushed before PM wake" logic (commit `fc47d0a`) — patching the same race at a call site instead of at the root.

Each of these was a real production bug fixed correctly *given the current design* — but they are all symptoms of the same missing invariant. New features keep paying this tax: any code path that persists task state must reason about invisible sibling instances.

## Proposed design

Make `activeTasks` the only source of `Task` objects:

1. **Private constructor, single factory.** `Task.get(taskId)` (and `Task.create(...)`) is the only way to obtain an instance. The factory checks `activeTasks` first, constructs at most once per process (behind the existing keyed lock), and always registers the instance before returning it. No caller — webhooks, recovery, merge checks, API routes — ever holds a Task that isn't the canonical one.
2. **All mutation flows through the instance.** External events become `task.sendMessage(...)` / `task.handleX(...)` calls on the canonical instance. Modules that today re-read `metadata.json` mid-flight (merge orchestrator, recovery idle checks) instead ask the live instance, which owns the in-memory truth and checkpoints to disk on its own schedule.
[...]

Issue #141, opened 2026-06-30
Run Archie on non-Anthropic / self-hosted models (model-swap spike)


egorkhmelev:
### What problem does this solve?

Archie is currently tied to Anthropic models — agents resolve to Claude aliases (`opus`, `sonnet`, …) and we lean heavily on the Claude Agent SDK. There is no way to run Archie on a third-party or self-hosted model today.

### Proposed solution

Be able to run Archie on a different underlying model — both third-party (e.g. DeepSeek, Codex/OpenAI) and self-hosted. Model-agnosticism opens up new options and is a potential selling point for the agent (not locked to a single vendor).

The hard, experimental part — and the real goal of this spike — is that we depend heavily on the Claude Agent SDK. We need to find out whether the model behind the SDK can be swapped for a different one and still have the whole system work. Deliverable: a feasibility finding (can we / can't we, and what it would take), surfacing the blockers that come from the SDK dependency. Distinct from the per-task pro-model upgrade in #142, which stays within Anthropic and the SDK.

### Alternatives considered

A translation/proxy layer that speaks the Anthropic API on one side and routes to other providers on the other is one possible approach; the mechanism is intentionally open at this stage.

### Area

Engine (orchestration / agent runtime / integrations)

### Pre-submission checks

- [x] I searched existing issues and this is not a duplicate.

egorshar:
Working on this from a fork — an **opencode-based agent runtime** to run Archie on non-Anthropic / self-hosted models, which is directly the SDK-swap feasibility this issue is after.

The foundation is **#209** (backend-seam extraction): it confines the Claude Agent SDK behind an `AgentRuntime` port (the `runtime/claude/sdk.ts` barrel + `ClaudeSdkRuntime`) and an `LlmOneShot` port, resolved via `getAgentRuntime()` / `getLlmOneShot()` — so a second runtime can slot in alongside `ClaudeSdkRuntime` without touching call sites.

On top of that seam I have an `OpencodeRuntime` running: an embedded `opencode serve` child per agent + an in-process tool bridge, so agents run on models like GLM / Grok via OpenRouter, with per-agent model routing and a per-child OS sandbox (bwrap + egress proxy on Linux). So the feasibility answer so far is **yes — the model behind the SDK can be swapped**, by going through opencode rather than the Claude Agent SDK directly.

[...]

PR #209, opened 2026-07-12
Pluggable repo host: backend-seam extraction + GitLab support

egorshar:
> First, thank you to the archie-hq maintainers — this is a fork contributing back. The codebase's clean separation across agents, tools, connectors, and tasks is exactly what made extracting these backend seams tractable rather than invasive, and we'd love to see multi-host support land upstream so the fork and origin don't drift. Grateful for the foundation to build on.

Makes Archie's repo host and agent runtime **pluggable behind ports**, then adds **GitLab** as a first-class repo host alongside GitHub. `REPO_HOST` defaults to `github` and the GitHub path is behaviorally unchanged — GitLab is fully opt-in via `REPO_HOST=gitlab`.

Two logical phases in one PR (commits are grouped and atomic if you'd prefer to review or land them separately):

### Phase 0 — backend seam extraction

Introduces a seam between Archie's business logic and its external backends so a second repo host or agent runtime can be added without touching call sites.

- New `src/ports/`: `RepoHost`, `RepoHostEventSource`, `AgentRuntime`, `LlmOneShot` interfaces + `RuntimeCapabilities`/`RepoHostCapabilities` descriptors, plus host-neutral domain types (`PRStatus`, `PRDetails`, `CheckRunReport`, `CodeScanningAlert`, …) relocated out of the GitHub connector so they're the canonical, vendor-neutral shapes.
- `src/system/backends.ts` — a single resolver that reads `REPO_HOST` / `AGENT_RUNTIME`, validates them at boot (fails fast with an actionable message), and hands callers a port-typed backend via `getRepoHost()` / `getAgentRuntime()` / `getLlmOneShot()`.
- `GitHubClient` now implements `RepoHost`; the Claude SDK is confined behind `runtime/claude/sdk.ts` (barrel), `ClaudeSdkRuntime`, and `ClaudeLlmOneShot`. Title generation, memory extraction/housekeeping, and triage now go through `LlmOneShot` instead of hand-rolling `query()`.
- Inbound webhook routing is extracted into a host-agnostic `connectors/shared/cr-router.ts` (with characterization tests); the merge orchestrator moves to `connectors/shared/merge.ts`.
- `GET /health` now reports the resolved `{ repoHost, runtime }` matrix.

### Phase 1 — GitLab repo host

A GitLab `RepoHost` + `RepoHostEventSource` that maps GitLab's REST/webhook surface onto the canonical GitHub-shaped types — an anti-corruption adapter, so agents, tools, and merge orchestration are unchanged.

- `src/connectors/gitlab/`: REST helper, `GitLabHost` client (MR read/list/details, create/update/close/comment/merge/push), CI reads (pipeline checks, job/pipeline-by-id, log tails), and webhook parse/verify/route into the canonical event vocabulary.
- Reviews synthesized from GitLab approvals + unresolved discussions; code-scanning alerts mapped from Ultimate-tier vulnerabilities (capability-gated).
- Capability probe at boot: `GET /license` raises `securityAlerts` only on Ultimate; defaults are least-capable otherwise.
- Host-aware clone URLs, git-askpass (`oauth2` + `GITLAB_TOKEN`), committer identity, and knowledge-log prefix (`gitlab:`), so cloning/pushing/logging resolve to the right host with no per-call-site branching.

### Compatibility

- Default is `REPO_HOST=github`; unset behaves exactly as today. No behavior change for existing GitHub deployments.
- `REPO_HOST=gitlab` requires `GITLAB_BASE_URL`, `GITLAB_TOKEN`, `GITLAB_WEBHOOK_SECRET` (validated at boot). See `docs/guides/gitlab-setup.md`.

### Testing

Unit tests for the ports, the resolver (github + gitlab), the GitLab client (read + write), status/ref mappers, and webhook parsing/routing. GitHub path covered by the existing suite, unchanged.

### Docs

- `docs/architecture/backends.md` — the seam, the resolver, and the capability model.
- `docs/guides/gitlab-setup.md` — operator setup + E2E checklist.

Thanks again for the foundation — happy to split this into stacked Phase 0 / Phase 1 PRs, trim scope, or iterate however suits your review process.


egorkhmelev:
Thank you for the contribution @egorshar. Given that this is big change, I will review it carefully and come back

src/system/__tests__/backends.test.ts
```diff
@@ -0,0 +1,36 @@
...
+    process.env.REPO_HOST = 'github';
+    expect(resolveRepoHostKind()).toBe('github');
+  });
+
+  it('rejects an unknown REPO_HOST value', () => {
+    process.env.REPO_HOST = 'bitbucket';
+    expect(() => assertBackendConfig()).toThrow(/REPO_HOST/);
+  });
+
+  it('rejects gitlab in phase 0 (not yet implemented)', () => {
```
egorkhmelev: **Must-fix (blocking).** This "rejects gitlab in phase 0 (not yet implemented)" case asserts the opposite of the contract this PR ships — `backends-gitlab.test.ts` proves `gitlab` is now an accepted host. It passes only coincidentally: with no `GITLAB_*` vars in the test env, the missing-env error message (`REPO_HOST=gitlab requires GITLAB_BASE_URL…`) happens to match the `/not available|gitlab/i` regex. It fails the moment CI exports `GITLAB_*`. Please delete it or rewrite it to assert the real contract (rejected with a missing-`GITLAB_*` message).

---

src/connectors/gitlab/webhooks.ts
```diff
@@ -0,0 +1,187 @@
...
+
+import crypto from 'crypto';
+import { extractTaskIdFromBranch } from '../github/branch-naming.js';
+import { findTaskByPRNumber, loadMetadata } from '../../tasks/persistence.js';
+import { determineRouteAction } from '../shared/cr-router.js';
+import type { NormalizedEventContext, RouteResult, RepoHostEventSource } from '../../ports/repo-host-events.js';
+
+/** Constant-time compare of the X-Gitlab-Token header against the configured secret. */
+export function verifyGitLabToken(token: string | undefined, secret: string): boolean {
+  if (!token || token.length !== secret.length) return false;
```
egorkhmelev: **Must-fix.** The length guard here compares JS string lengths (`token.length !== secret.length`), but `crypto.timingSafeEqual` requires equal **byte** lengths and throws `RangeError` otherwise. A token with the same UTF-16 character count as the secret but a different UTF-8 byte length (any multibyte character) slips past the guard and makes `timingSafeEqual` throw — e.g. `verifyGitLabToken('sécret', 'secret')`. This is pre-auth and remotely triggerable on `POST /webhooks/gitlab`; under Express 5 the async rejection becomes a 500 instead of a 401, and the 401-vs-500 flip is an oracle for the secret's character length. Compare `Buffer.byteLength(token) !== Buffer.byteLength(secret)` (or hash both sides before `timingSafeEqual`).

---

src/connectors/shared/cr-router.ts
```diff
@@ -0,0 +1,155 @@
...
+ *
+ * Called for: PR approval, push, CI success
+ * Debounces to avoid redundant checks when webhooks arrive in bursts.
+ */
+export function handleMergeCheckDirect(taskId: string): void {
+  // Cancel any pending merge check for this task
+  const existingTimer = mergeCheckTimers.get(taskId);
+  if (existingTimer) {
+    clearTimeout(existingTimer);
+    logger.system(`GitHub: Debouncing merge check for task ${taskId}`);
```
egorkhmelev: If this router stays shared across hosts, these debounce/fire log lines still hardcode `GitHub:` (here plus lines 43, 76, 81). They execute on the GitLab webhook path too (`gitlab/events.ts` → `handleMergeCheckDirect`/`handleChecksReadyDirect`), so a GitLab deployment's logs would claim GitHub activity — contrary to this PR's own host-agnostic intent and its `push_branch` "Repo: Pushed…" neutralization. Neutralize to `Repo:`/`cr-router:` or make it host-correct. (Non-blocking, but note this whole shared-relocation is part of what's under discussion in the scope re-scope above — if the merge/router work is deferred, this goes with it.)

---

src/tasks/__tests__/title-generator.test.ts
```diff
@@ -202,16 +220,18 @@ describe('generateTaskTitle', () => {
       ],
     });
     await generateTaskTitle(thread);
-    const transcript = state.lastQueryArgs.prompt as string;
+    const transcript = state.jsonMock.mock.calls[0][0].prompt as string;
     expect(transcript).toContain('[forwarded from @<UEXT:External> — external, team T_OTHER]');
```
egorkhmelev: **Must-fix (merge conflict).** The branch is conflicting with `main` — `git merge-tree` reproduces exactly one conflict, here in this file. This side still expects the pre-flip marker `[forwarded from @<UEXT:External> …]`, but main's `c3d6f35` flipped the producer to the Slack-native form `<@UEXT:External>`. Resolve by **combining both sides**: keep this PR's `state.jsonMock.mock.calls[0][0].prompt` plumbing **and** adopt main's `[forwarded from <@UEXT:External> — …]` expected string. A verbatim PR-side resolution fails the merged suite (verified). The adjacent `appendGitHubEvent` source label auto-merges cleanly, so no action needed there.

---

egorkhmelev (changes requested):
## Archie/Forge review — request changes

_This is an automated review by Archie/Forge, run independently of the author. It derives intent and acceptance criteria from the PR itself, then runs review and QA rings in an isolated worktree. Flagging assumptions where it made them._

Thanks for this — the GitLab adapter itself is careful work, and the compatibility discipline on the GitHub path shows. The ask here is about **scope**, not quality.

### Please isolate this PR to GitLab support only

Split out — into a separate proposal, not this PR — the agent-runtime and LLM-vendor abstraction: the `AgentRuntime` and `LlmOneShot` ports and the "confine the SDK to `src/runtime/claude/`" vendor-isolation gate. Two reasons:

1. **No usecase yet.** These are groundwork for a future swappable runtime (#141), but there's no second runtime in the tree, so the abstraction can't be exercised or verified today — it's complexity paid up front for a direction we're not ready to commit to. How agents should be abstracted is still R&D on our side; we don't want to lock in a specific solution before we've done that work.
2. **It carries real risk for zero GitLab benefit.** The one-shot consolidation quietly changes error logging on the GitHub path (extractor error subtypes are swallowed; triage's two distinct error paths collapse into one generic message), which undercuts the "behaviour unchanged" claim. None of it is needed to run GitLab.

Keeping **just the repo-host seam** — the `RepoHost` interface + env resolver + GitLab webhook source — is the right size: it has a concrete second consumer (GitLab), so the abstraction earns its place. Please also have it reshape the existing GitHub path as little as possible — e.g. relocating the merge orchestrator to `shared/` is what forced a GitHub-path test to move; prefer passing the host in over moving files.

### Direction this fits into

We're actively pursuing a simplification pass that this intersects: **#202** (move GitHub *reads* to `gh` + a per-agent scoped read-only token in the sandbox, deleting ~15 read-only tool wrappers) and **#203** (Slack as a pure event subscriber), both alongside **#201** (single canonical Task instance). #202 in particular means much of a repo host's *read* surface is about to disappear — so the GitLab adapter should lean thin and avoid reimplementing reads we'd soon delete. The discussions→review-thread synthesis and the Ultimate-tier vulnerabilities mapping are the clearest candidates: under #202 a GitLab agent would just run the API directly rather than go through a hand-maintained mapping (which also happens to be the part we can't verify upstream). The guiding principle is the one the whole system is built on: abstract where there's a usecase, stay dumb-simple everywhere else.

### Must-fix regardless of scope

Inline comments on each:
- **`src/system/__tests__/backends.test.ts`** — the "rejects gitlab in phase 0" test asserts the opposite of the shipped contract and passes only coincidentally (blocking).
- **`src/tasks/__tests__/title-generator.test.ts`** — branch conflicts with `main`; resolve by combining the PR's mock plumbing with main's `<@UEXT:External>` expected string.
- **`src/connectors/gitlab/webhooks.ts`** — webhook token check compares JS string length before `timingSafeEqual` (needs byte length) → `RangeError`/500 and a length oracle.
- **`src/connectors/shared/cr-router.ts`** — if the shared router stays, its debounce/fire logs still say `GitHub:` and misreport under GitLab.

### Couldn't verify upstream

The live GitLab clone/commit/push/MR flow and the Ultimate-tier code-scanning API shapes rely on a GitLab instance we don't have — we're treating those as author-attested, not independently confirmed.

Happy to look again once it's re-scoped. Appreciate the work here — the GitLab half is worth landing on its own.

---

egorshar:
## Re-scoped to GitLab repo-host only + live E2E

Thanks for the review — I've re-scoped the PR to address the feedback. Summary of what changed and how it maps to your asks:

### 1. Isolated to the repo-host seam (agent-runtime / LLM abstraction removed)
- Removed the `AgentRuntime` and `LlmOneShot` ports and the "confine the SDK to `runtime/claude/`" vendor-isolation from this PR. They'll come back as a **separate proposal** tied to the future swappable-runtime work (#141), not bundled here.
- Reverted the `triage` / `memory` / `title` one-shot refactors to direct SDK calls. This also **removes the GitHub-path error-logging change** you flagged — it was a side effect of the one-shot consolidation, so reverting eliminates it rather than patching it.

### 2. GitHub path left essentially untouched; merge/router deferred
- The merge-on-green orchestrator was **not** relocated and the shared `cr-router` extraction is gone. Concretely: `git diff <base> -- src/connectors/github/webhooks.ts` and `-- src/connectors/github/merge.ts` are both **empty**, and there is no `src/connectors/shared/cr-router.ts`.
- The only `github/` + `shared/` changes are the seam itself: `github/client.ts` (`implements RepoHost` + result-type relocation to `ports/`), host-aware clone, and `shared/repo-url.ts`.
- GitLab webhooks route only to "wake the existing task" or "discard" — **no auto-merge orchestration** on the GitLab side. `RepoHost.mergePullRequest` and the `merge_pull_request` tool are retained (interface contract); only the webhook-driven orchestrator is deferred.

### 3. Reads kept thin
- Dropped the Ultimate-tier vulnerabilities→code-scanning mapping (the unverifiable part). GitLab's `securityAlerts` capability is fixed `false`, so those tools simply aren't surfaced there.

### 4. Must-fixes
- **`backends.test.ts`** — rewritten to assert the real contract (`gitlab` rejected only when the `GITLAB_*` env is unconfigured; resolves when present).
- **GitLab webhook token check** — now compares UTF-8 **byte** length before `timingSafeEqual` (with a multibyte regression test), so a multibyte token can't throw `RangeError` → 500, and there's no length oracle.
- **title-generator conflict** and the **`cr-router` `GitHub:` logs** both dissolved under the re-scope (title-generator is untouched from `main`; there is no `cr-router` here).

### One deviation I'd like your call on
I **kept** the discussions→review-thread **synthesis**. It's ~50 lines entirely within the GitLab client's `getReviews`, with **zero** GitHub-path footprint (GitHub's `getReviews` is byte-identical to `main`) and no dependency on the deferred merge/router. GitLab has no native review-state object, so until the read surface is reworked, this is what gives a GitLab agent real review feedback (approvals + change-requests). Happy to defer it too if you'd prefer — flagging rather than silently keeping it.

### Live end-to-end
Because CI can't reach a VPN-internal GitLab, I ran the branch against a real self-hosted GitLab instance and attached the results below (author-attested). All the integration paths this PR adds — git-askpass HTTPS clone/fetch/push, the full agent MR flow, signed-webhook receipt with byte-safe token verification, event normalization, and webhook→task re-activation (a new agent turn) — passed. Report below.

<details>
<summary><b>Live E2E report</b> (self-hosted GitLab; placeholders used for private host/repo/user)</summary>

**Branch under test:** the re-scoped GitLab repo-host branch @ `<sha>`
**Target:** a self-hosted GitLab instance (`<gitlab-host>`) + a test project `<group>/<repo>`
**Method:** the branch was booted in Docker from the checkout under test and driven end-to-end against the live instance. CI cannot reach a VPN-internal GitLab, so this is an **author-attested live run**; the harness attests the running image was composed from the branch SHA.

#### Verdict
Boot + host resolution, git-askpass HTTPS clone/fetch/push, the full agent MR flow (`createPullRequest`), signed-webhook receipt with byte-safe token verification, event parse/normalization, and webhook → task re-activation (new agent turn) incl. an MR reply comment — all pass. No defects in the code under test.

#### Environment / attestation
- `/health`: `{"status":"ok","git_sha":"<sha>","backends":{"repoHost":"gitlab"}}`; harness `Attested: instance composed from <sha>`.
- `REPO_HOST=gitlab`; startup log `Backends: repoHost=gitlab`.
- Transport note: the deployment's *other* repos use SSH remotes; their in-container startup pulls fail `Permission denied (publickey)` (no SSH key in the container) and fall back to cached clones — expected/unrelated. The **test repo uses an HTTPS remote**, which is exactly the `git-askpass` path this PR adds, so it was used for the run.

#### Results
| # | Item | Result | Evidence |
|---|------|--------|----------|
| 1 | Boot from branch; `repoHost=gitlab`; `/health` attests SHA | ✅ | `/health` + `Attested:` |
| 2 | git-askpass HTTPS clone/fetch | ✅ | startup pull of test repo succeeds (no auth error); agent `Fetched origin` |
| 3 | git-askpass HTTPS **push** | ✅ | direct probe: `[new branch] … PUSH_EXIT=0`, then deleted |
| 4 | Edit-mode approval gate | ✅ | `approval_requested (edit_mode)` → approved |
| 5 | Agent: clone → edit → commit → push → **open MR** (`createPullRequest` REST) | ✅ | MR !N, 1 file / 1 insertion, not merged |
| 6 | MR state via REST (adapter read path) | ✅ | `iid:N state:opened detailed_merge_status:mergeable target:main` |
| 7 | Webhook signed receive → 200; bad token → 401 (byte-safe verify) | ✅ | `HTTP=200 {"received":true}` / `bad-token HTTP=401` |
| 8 | Webhook parse/normalize (push, pipeline→workflow_run, note→issue_comment) | ✅ | normalized log lines |
| 9 | Webhook → resolve task → **wake → new turn** | ✅ | `task:resumed` + PM new turn reading the comment |
| 10 | MR-comment write (agent reply) | ✅ | reply note present via REST |
| 11 | Committer identity, shared clone, webhook mount | ✅ | logs |

Not driven live (covered by the 928-test unit suite): review-synthesis state mapping, code-scanning capability-gate short-circuit, status/ref mappers, merge-via-tool, conflict mapping.

#### Evidence detail

git-askpass HTTPS push (direct probe, in-container):
```
REPO_HOST=gitlab  GIT_ASKPASS=/app/scripts/git-askpass.sh
remote: https://<gitlab-host>/<group>/<repo>.git
 * [new branch]      <probe-branch> -> <probe-branch>
PUSH_EXIT=0
 - [deleted]         <probe-branch>
DELETE_EXIT=0
```

Agent MR flow (completion log) + REST confirmation:
```
[<repo>-agent] [completion] created branch `archie/task-<id>` off main, added a one-line
comment to README.md (single-line diff), committed, pushed, opened MR !N into main WITHOUT merging.
```
```
"iid":N  "state":"opened"  "target_branch":"main"
"source_branch":"archie/task-<id>"  "detailed_merge_status":"mergeable"
```

Webhook token verification + parse:
```
# correct X-Gitlab-Token → accepted
HTTP=200  body: {"received":true}
# wrong X-Gitlab-Token → rejected (byte-safe compare before timingSafeEqual; no 500)
bad-token HTTP=401
```
```
GitLab webhook: push/ repo=<group>/<repo> branch=<probe-branch>
GitLab webhook: workflow_run/pending … (GitLab pipeline → canonical workflow_run)
GitLab webhook: workflow_run/completed … state=failure
GitLab webhook: issue_comment/created repo=<group>/<repo> mr=!N branch=archie/task-<id> user=<reviewer>
```

Webhook → task re-activation (new turn): a correctly-signed `note` for the MR's branch (sender ≠ bot user, so not self-filtered) re-activated the completed task:
```
message: {"from":"<reviewer>","to":"pm-agent","destination":"<group>/<repo>/MR !N",
          "message":"… is the marker line intentional? …"}
task:resumed: {}
[pm-agent] agent:active: {}
pm-agent → <repo>-agent: "Follow-up on MR !N … the reviewer left a comment … please post a reply …"
```
The new turn completed and posted a reply comment; REST confirmed the note on the MR.

</details>

src/agents/tools.ts
egorkhmelev, line 1: Non-blocking follow-up (host-correctness). The push tool around line 1100 hardcodes `logger.system(\`GitHub: ${message}\`)`, but push is host-agnostic (plain git), so under `REPO_HOST=gitlab` this log line claims "GitHub:" activity on a GitLab deployment. This is the same class as the round-1 cr-router finding (which the re-scope revert resolved) — this instance is a pre-existing line the diff didn't touch, now exposed by adding GitLab. Neutralize the prefix (e.g. `Repo:`) or make it host-aware. Fine to handle as a follow-up.

---

src/system/__tests__/backends.test.ts
```diff
@@ -0,0 +1,58 @@
+import { describe, it, expect, afterEach, vi } from 'vitest';
+
+// The GitLab connector client lands in a later task and doesn't exist on disk
+// yet, so isolate backends.ts's import of it here. NOTE: this specifier must
+// stay textually identical to the import in backends.ts (not re-derived from
+// this file's own relative path) — since the target module doesn't resolve to
+// a real file, Vitest's mock matching falls back to raw specifier text rather
+// than a canonicalized path.
+vi.mock('../connectors/gitlab/client.js', () => ({
```
egorkhmelev, line 9: Non-blocking cleanup. This `vi.mock` is now both stale and ineffective: the comment says "the GitLab connector client … doesn't exist on disk yet", but `src/connectors/gitlab/client.ts` ships in this same PR. And the specifier `'../connectors/gitlab/client.js'` resolves relative to this test file (`src/system/__tests__/`) to `src/system/connectors/gitlab/client.js`, which doesn't exist — so it never intercepts `backends.ts`'s import of the real client. It's a no-op mock with a misleading comment; either drop it or fix the path + comment if the interception was actually intended.

---

src/connectors/gitlab/webhooks.ts
```diff
@@ -0,0 +1,232 @@
...
+export function classifyGitLabEvent(context: NormalizedEventContext): 'existing_task' | 'discard' {
+  const { eventType, action, state } = context;
+
+  switch (eventType) {
+    case 'pull_request':
+      if (action === 'closed' || action === 'opened' || action === 'synchronize') return 'existing_task';
+      return 'discard';
+
+    case 'pull_request_review':
+      if (state === 'approved' || state === 'changes_requested' || state === 'commented') return 'existing_task';
```
egorkhmelev, line 167: Non-blocking simplification. This gates `pull_request_review` on `state === 'changes_requested' || state === 'commented'`, but `formatGitLabContext` only ever emits a `pull_request_review` with state `'approved'` (GitLab's `approved` MR action) — change-requests arrive as note/discussion events and are classified elsewhere. So both disjuncts here are unreachable; the check can be reduced to `state === 'approved'`.

---

egorkhmelev (approved):
## Archie/Forge review (round 2) — approve with comments

_Automated follow-up review by Archie/Forge, run independently against the current head (`5117054`) after the re-scope. Delta-reviewed against the previously reviewed head and re-ran the QA ring._

The re-scope is exactly what was asked for, and it holds up under independent inspection — this supersedes my earlier change request. The agent-runtime / LLM-vendor abstraction is genuinely gone: `AgentRuntime`/`LlmOneShot`, the `runtime/claude/` SDK isolation, the title/memory/triage one-shot refactors, the shared `cr-router` extraction, and the merge-orchestrator relocation are all reverted — `title-generator.ts`, `extractor.ts`, `triage.ts`, and `merge.ts` are byte-identical to `main`. What ships is the `RepoHost` + `RepoHostEventSource` seam + GitLab, plus the security hardening. Thank you for the thorough turnaround.

**All four round-1 must-fixes are resolved:** the stale phase-0 test now asserts the real contract; the merge conflict is gone (clean `merge-tree` against main); the webhook token check compares UTF-8 byte length before `timingSafeEqual`; the code-scanning path is reduced to a capability-gated stub that makes no API call. QA re-ran green: 939 unit tests + `tsc` clean, GitHub-default and `REPO_HOST=gitlab` resolution both verified, and the git-askpass host-identity split checked live.

Approving so this can land. The items below are all **non-blocking follow-ups** — none are architectural:

- **`src/agents/tools.ts` (~line 1100)** — the host-agnostic push tool still logs a hardcoded `GitHub:` prefix, so a GitLab deployment's push logs misreport the host. Same class as the round-1 cr-router finding; this is a pre-existing line exposed by adding GitLab. (Inline note.)
- **`src/system/__tests__/backends.test.ts:9`** — a `vi.mock` that's now stale (its comment claims the GitLab client doesn't exist yet, but it ships here) and ineffective (the specifier resolves to a nonexistent path, so it never intercepts). (Inline note.)
- **`src/connectors/gitlab/webhooks.ts:167`** — `classifyGitLabEvent` gates review events on `changes_requested`/`commented`, which `formatGitLabContext` never emits; reducible to `approved`. (Inline note.)
- A few smaller dead-code residues remain (the unreachable `else` in `index.ts:244`, the unused `gitlabEventSource` conformer, `setCapabilities()` / `repoEventPrefix()` with no callers) — worth a cleanup pass in the same spirit.
- **Please update the PR description** — it still narrates the original two-phase design including the Phase-0 abstraction that was reverted, so it no longer matches what ships.

**Not independently verified:** the live GitLab clone/commit/push/merge flow against a real instance (no reachable GitLab here) — treating your end-to-end run as author-attested. The `GitLabHost.getPRReviews` discussions→review synthesis is left in as you flagged it pending a maintainer call; fine to keep.

Nice work getting this to a focused, single-purpose change.

---

egorshar:
@egorkhmelev The branch is all green now — approved, all four checks passing (typecheck/build/test, CodeQL, secret scan, JS/TS analysis), and no conflicts with `main`. Do you plan to merge it?

One note: after your approval, CodeQL posted a non-blocking "missing rate limiting" alert on the GitLab webhook route (`src/connectors/gitlab/events.ts:46`). The CodeQL check itself passes, so it doesn't block the merge — happy to address it in a follow-up PR alongside the other non-blocking items, or add a limiter now if you'd prefer it in this PR.

Issue #153, opened 2026-06-30
Per-user claude.ai subscription auth for DMs (cost-reduction experiment)


egorkhmelev:
### Goal

Let DM interactions run on each user's **own claude.ai subscription** instead of shared org API credits. Each user authenticates their own Claude account once, and Archie reuses that auth to serve that user's DM tasks. The point is lower cost — a flat-rate personal subscription instead of metered API usage — for personal DM work.

### What to explore

- Per-user subscription auth in the Claude Agent SDK (it supports subscription-based login as an alternative to an API key).
- Securely associating a user's auth with their identity and using it **only** for that user's own tasks (their DMs).
- Routing: DM tasks use the requesting user's subscription; shared/channel work stays on the default credentials.
- Feasibility, the UX of the one-time setup, and the actual cost delta.

### Notes

- Confirm this is within Anthropic's terms for subscription-based authentication (per-user, user-consented, that user's own usage) before relying on it.

### Area

Engine (orchestration / agent runtime / integrations)

### Pre-submission checks

- [x] I searched existing issues and this is not a duplicate.

PR #226, opened 2026-07-19
feat(auth): Claude subscription auth for CLI-only interactive use

egorshar:
Lets an individual running Archie in **CLI-only mode** authenticate the spawned Claude Code CLI/SDK with their Claude subscription (`CLAUDE_CODE_OAUTH_TOKEN`, from `claude setup-token`) instead of a metered `ANTHROPIC_API_KEY` — the same subscription-auth path Claude Code itself supports for Pro/Max plans.

## What it does
- **`resolveClaudeCredential()`** (`src/system/claude-credential.ts`) auto-detects, in priority order: `ANTHROPIC_API_KEY` → `CLAUDE_CODE_OAUTH_TOKEN` → none, and normalizes the result to the env fragment the spawned CLI honors. Token values are never logged — only the resolved kind (`api_key` / `oauth_token_env`, logged as `Claude auth: api_key` / `Claude auth: oauth_token (env)`).
- **Boot gates on *any* Claude credential** (`assertClaudeCredentialAvailable()`), replacing the previous `ANTHROPIC_API_KEY`-only check, with a clean, actionable `StartupError` (`src/system/startup-error.ts`) — no JS stack trace — when none is present. Headline **and** details print to stderr, so `2>startup.log` captures both.
- **The resolved credential is threaded into every Claude SDK env allowlist** — agent spawn plus the one-shot calls (title generation, memory extractor/housekeeping, triage, research) — via `claudeCredentialEnv()`.

Default behavior is unchanged: an `ANTHROPIC_API_KEY`-configured deployment resolves exactly as before.

## Scope — CLI-only interactive use
Subscription auth is for **CLI-only interactive use** (`npm run cli`) — turns are human-initiated, the same shape as using Claude Code directly. **Slack-connected deployments and the E2E harness are automated access and use `ANTHROPIC_API_KEY`.** The E2E preflight (`tools/e2e/boot.ts`) requires the API key deliberately. Documented as a one-line scoping note in `.env.example` and `docs/guides/local-development.md` (no licensing prose in the config template); `docs/guides/deployment.md` corrected.

## Follow-ups (not this PR)
- #153 — this doesn't close it.
- The `npm run cli` self-bootstrap (auto-`claude setup-token` + auto-started dev server) — the natural home for this feature given the CLI-only lane.

## Verification
`npm run typecheck` ✓ · `npm run build` ✓ · `npx vitest run` → 914/914 ✓ (includes `claude-credential` unit tests pinning the label and the startup-error env-var names).

https://claude.ai/code/session_0149B53Y2e1THfaDsNbrqvQc

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,5 +1,13 @@
-# Anthropic API
+# Claude authentication (auto-detected in priority order):
+#   1. ANTHROPIC_API_KEY — standard API key.
+#   2. CLAUDE_CODE_OAUTH_TOKEN — subscription token from `claude setup-token`.
+# At least one must be available or startup fails.
+#
+# Subscription auth (CLAUDE_CODE_OAUTH_TOKEN) is for CLI-only interactive use
+# (`npm run cli`); use ANTHROPIC_API_KEY for Slack-connected deployments and the
+# E2E harness. See docs/guides/local-development.md.
 ANTHROPIC_API_KEY=sk-ant-...
+# CLAUDE_CODE_OAUTH_TOKEN=
 
 # Claude Code CLI path - point directly to cli.js
 CLAUDE_PATH=/path/to/claude
--- a/docs/guides/deployment.md
+++ b/docs/guides/deployment.md
@@ -27,7 +27,7 @@ Monitoring: container logs + /health endpoint
 Secrets are injected via the container's environment file plus the mounted
 `/app/secrets` volume. See `.env.example` for the full list. Required at runtime:
 
-- `ANTHROPIC_API_KEY` — Claude API access (required; startup fails without it)
+- `ANTHROPIC_API_KEY` — Claude API access. A deployment must use an API key (subscription auth via `CLAUDE_CODE_OAUTH_TOKEN` is CLI-only). Startup fails if no Claude credential is present.
 - `SLACK_BOT_TOKEN` / `SLACK_SIGNING_SECRET` — Slack integration in HTTP webhook mode (optional; CLI-only mode if both omitted)
 - `SLACK_APP_TOKEN` — `xapp-...` app-level token; set this *instead of* `SLACK_SIGNING_SECRET` to use Socket Mode and deploy without an inbound webhook URL
 - `GITHUB_APP_ID`, `GITHUB_APP_SLUG`, `GITHUB_INSTALLATION_ID` — GitHub App identifiers
--- a/docs/guides/local-development.md
+++ b/docs/guides/local-development.md
@@ -2,12 +2,14 @@
 
 ## Quick Start
 
-The only required env var is `ANTHROPIC_API_KEY`. Slack and GitHub App credentials are optional.
+Archie needs a Claude credential, auto-detected in priority order: `ANTHROPIC_API_KEY`, then `CLAUDE_CODE_OAUTH_TOKEN` (a subscription token from `claude setup-token`). At least one must be set or startup fails. Slack and GitHub App credentials are optional.
+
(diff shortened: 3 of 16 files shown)


egorkhmelev (commented):
Thanks, and sorry it sat. Shape is right — resolver module spread into the existing `env:` allowlists, no new abstraction, additive for API-key deploys. All six call sites covered, no token value logged. Happy to take this with a few small things.

**Scope it to CLI-only mode.** The lane for subscription auth is the CLI-only server (no Slack credentials) driven from `npm run cli` — human-initiated turns, same shape as using Claude Code directly. Slack-connected deployments and the E2E harness are both automated access and should stay on `ANTHROPIC_API_KEY`. `tools/e2e/boot.ts:52` requiring an API key is therefore correct as-is — no change needed, but worth a sentence saying it's deliberate. Right now the docs land in `local-development.md`, which covers `npm run dev` too, so it reads as blessing the Slack-connected server.

**That replaces the Terms prose.** `.env.example:6-12` and the guide currently commit a verbatim Consumer Terms quote plus a legal conclusion. I'd rather not carry a licensing position in a config template — it goes stale and isn't ours to settle. One scoping line does the job better: "Subscription auth is for CLI-only interactive use; use `ANTHROPIC_API_KEY` for Slack-connected deployments and the E2E harness."

**Small fixes:**

- `src/index.ts:291-297` — `StartupError` puts the headline on stderr (`logger.error`) and `details` on stdout (`logger.plain`), so `npm run dev 2>startup.log` gets the problem without the fix. Emit both on stderr.
- `docs/guides/deployment.md:30` says `ANTHROPIC_API_KEY` is "required; startup fails without it" — no longer true.
- `local-development.md` still says set `ANTHROPIC_API_KEY` (line 10) and lists only it under `# Required` (143-145); the scoping above should sort this out.
- Tests don't assert the `'oauth_token (env)'` label or that `StartupError.details` names both vars — worth pinning, they're the operator-facing strings. (Also the PR body says `oauth_token_env` where the code emits `oauth_token (env)`, and `KIND_LABEL.none` is unreachable.)

Please rebase — main moved the SDK `^0.3.202` → `^0.3.219` and touched `spawn.ts` after your 838/838.

Follow-ups, not this PR: #153, and the `npm run cli` self-bootstrap you floated — given CLI-only is the lane, that's where this feature really wants to live. Agreed this doesn't close #153, and thanks for saying so rather than claiming it.

egorshar:
Rebased onto current main (SDK `0.3.220`, `spawn.ts` reapplied cleanly). Changes:

- **Scoped to CLI-only.** Docs now state subscription auth is for CLI-only interactive use (`npm run cli`); Slack deployments and the E2E harness use `ANTHROPIC_API_KEY`. Dropped the Consumer-Terms quote/conclusion from `.env.example` for the one-line scoping note. `local-development.md` reframed (Quick Start + `# Required` list); `deployment.md`'s "required; startup fails" corrected. Added a comment at `tools/e2e/boot.ts` noting the API-key requirement is deliberate (no behavior change — it stays on an API key).
- **StartupError** now emits headline and details on stderr (new `logger.plainErr`), so `2>startup.log` captures both.
- **Tests** pin the `oauth_token (env)` label and that the startup error names both env vars. (You're right on the naming: the *kind* is `oauth_token_env`, the logged *label* is `oauth_token (env)` — I mislabeled it in the PR body and have corrected it. `KIND_LABEL.none` is unreachable since the guard throws first; left in the map for completeness.)
- Follow-ups noted, not this PR: #153 stays open, and the `npm run cli` self-bootstrap is the natural home for this given the CLI-only lane.

PR #244, opened 2026-07-29
fix(triggers): let the agent manage a proposal awaiting approval

egorkhmelev:
> Bottom of a three-PR stack (#244 → #245 → #246); merge this first. Commit SHAs are deliberately not cited — the stack is rebased as it moves, so any SHA here would dangle.

Makes a trigger proposal that is still `pending` a first-class, manageable object — and makes the multi-card, mutable-while-pending world that creates safe.

## Problem

Pending was a dead zone: `list_triggers` filtered `status !== 'pending'` out, and `update_trigger` answered a pending id with **"No trigger `<id>` found."** So when a user neither approved nor denied a proposal and then asked for a change, the agent was told the revision target didn't exist. Its only remaining move was to propose a *second* trigger and abandon the first — still on disk, absent from every listing, with a live Approve button able to enable a superseded automation later.

Observed in task-20260729-1054-e0dc5w: `trg-20260729-1056-6nd69p` left pending beside the enabled `trg-20260729-1057-rdvrxx`, identical cron and channel. `delete_trigger` would in fact have accepted a pending id, but its description sourced ids from `list_triggers`, which never showed it — the one escape hatch was unreachable by design.

## What it does

Pending proposals are listed (`awaiting approval — not running`), editable in place (re-posting the Approve/Deny card under the **same** trigger id so the user decides on current details), and withdrawable. Card construction is shared between propose and edit. Two invariants hold throughout: approval stays the user's exclusive act — a `status` change on a proposal is refused on every surface — and nothing is announced to the bound channel, because the automation isn't running.

Cards are deliberately **never retracted**, so multiple live cards for one trigger are normal. That is absorbed by making both resolutions idempotent by status rather than by rewriting Slack messages.

## Consequences that had to be handled

- **Management is scoped to the proposing task** (`Trigger.proposed_in_task`), not the binding. Binding visibility is wrong in both directions: `propose_trigger` applies no visibility check, so a DM can bind a proposal to a *private* channel that binding rules then hide from that same DM — the dead zone reappearing in its likeliest case — while a *public*-bound proposal would be editable **and approvable** from any other conversation, going live with a second user's content while `created_by` still names the requester. Pre-existing proposals fall back to binding visibility; operator context sees everything.
- **A stale Deny can no longer tear down a live trigger.** Approval already tolerated a second click (`enableProposedTrigger` requires `pending`); denial deleted whatever id it was handed — and deleted only the *file*, never calling `deindexTrigger`, so the scheduler kept firing a ghost until restart while the card said "Declined" regardless of outcome. Denial is gated on `pending` and returns a semantic outcome; a refusal is recorded in the task's knowledge log.
- **Editing is a true read-modify-write.** It re-reads immediately before writing and applies the edit to *that* object, so neither a concurrent Approve (which would be undone) nor a concurrent content edit (which would be lost) survives unnoticed.
- **Editing renews the GC clock.** The boot scan reaps proposals older than 24h; measuring from `created_at`, which an edit didn't refresh, reaped a proposal revised 23h in — taking its just-posted card with it. `pendingSince` measures from `updated_at` when set, on both the agent and operator paths. Abandoned proposals still age out.
- **A cap hit at approval leaves the proposal pending.** It used to delete the file, making the refusal indistinguishable from a withdrawal: the card said "no longer around", never that a cap was the reason, and the proposal was gone so freeing a slot didn't help.
- **Every surface reports the same truth.** Card copy reads the trigger's real state instead of a two-cause warning for what is now routine; the operator API stops hiding pending proposals (that folder documents full operator visibility with no bypass) and answers `409 { stale: true }` when a resolution changed nothing, matching the merge gate; the CLI list gives pending its own glyph and explains why `[p]` doesn't apply, instead of surfacing a raw `409`.

Explicit non-goal: cap accounting. Orphaned proposals never consumed caps — `countActiveTriggers` counts `enabled` only.

## Verification

Typecheck clean. Suite **1012/1012** at the head of the stack (955 on this branch alone).

| File | Tests |
|---|---|
| `src/agents/__tests__/trigger-tools.test.ts` | 21 |
| `src/tasks/__tests__/trigger-denial.test.ts` | 9 |
| `src/connectors/__tests__/trigger-operator-surface.test.ts` | 11 |
| `src/system/__tests__/trigger-pending-ttl.test.ts` | 4 |
| `src/cli/components/TriggerList.test.tsx` | 6 |

Eleven mutations across two rounds — dropped ownership gate, dropped delete gate, dropped race re-check, stale-object save, dropped `updated_at`, reaper ignoring `updated_at`, refused-Deny clearing state, cap deletion, missing CLI status case, CLI PATCH-on-pending, dropped operator race guard — each fail at least one test.

Two test harnesses did not exist before this PR (`TriggerList.test.tsx` for the ink TUI, `trigger-operator-surface.test.ts` driving the mounted Express router), which is why the CLI and operator-path defects got through the first round.

## Live verification

A booted instance confirmed the point of the change: given a revision request on an unapproved proposal, the PM replied *"Updated the same proposal — no second automation was created"* and pointed at the fresh prompt. Single trigger file on disk.


egorkhmelev:
## Added: stale-Deny hardening (d8c3b40)

Reviewing what happens to the **earlier approval card** after an edit re-posts a new one turned up an adjacent bug worth fixing here, since this PR makes duplicate cards intentional.

Cards are never retracted, so an older card stays clickable after its trigger is approved. Approval already tolerated that — `enableProposedTrigger` requires `pending`, so a second Approve is a no-op. **Denial did not:** `handleTriggerDenial` deleted whatever id it was handed, with no status check.

So a Deny on a stale card **destroyed a live automation** — and only deleted the *file*: `deindexTrigger` is never called on that path, so the scheduler kept firing it from the in-memory index until the next restart, then it vanished. The Slack card also wrote "Declined" *before* the handler ran, so it said that regardless of outcome.

- `handleTriggerDenial` only deletes a still-`pending` proposal; returns `withdrawn` / `already_live` / `not_found`.
- The Slack handler resolves first, then renders the real outcome — pointing at pause/remove when the proposal was already approved.
- Same gate on the direct-store fallback (no task found for the thread).
- Skill documents the card semantics so the agent tells the user which card is current.

`src/tasks/__tests__/trigger-denial.test.ts` — 7 tests, mutation-checked (removing the status gate fails 2). Suite now 922/922 on this branch.

PR #246, opened 2026-07-29
feat(reminders): recurring reminders via cron, so the runtime owns the cadence

egorkhmelev:
> Stacked on #245 (base `fix/reminder-trigger-guidance`), which is stacked on #244. Commit SHAs are deliberately not cited — this branch is rebased as the stack moves, so any SHA here would dangle.

Moves ownership of a repeating reminder's cadence out of the agent and into the runtime.

## Problem

Recurring monitoring was the *agent's* job: wake, decide "in 12 hours", call `set_reminder` again. Every wake re-derived the interval. Measured across all 40 re-arms of **task-20260617-1454-i1a08v** (a 12-hourly offer-code monitor running since 2026-06-17):

| Symptom | Evidence |
|---|---|
| Wrong interval | **6 of 40** — 5× 24h, 1× 22.7h — silently halving the monitoring rate |
| Slot drift | **37 min** (05:15 → 05:52 over 20 days): each re-arm was "12h from *this wake*" |
| Missed checks | **36 fires** in a 481h span that owes 40 at a true 12h cadence |
| Config-as-string | cadence lived inside a **~1900-char `reason`**, mixed with the monitor's state |

## What it does

`set_reminder` takes `cron` + `tz` (+ optional `until`) as a second mode alongside one-shot `datetime`. The scheduler re-arms from the cron after each fire instead of clearing, so the runtime owns the cadence and it cannot drift, be skipped, or be misremembered. Reuses the trigger scheduler's `computeNextRun` / `validateRecurringInterval` rather than adding a second recurrence engine — reminders shipped without cron because that infrastructure didn't exist yet.

- `planReminderRearm` is the pure rule: one-shots clear; recurring advance to the next cron instant strictly after the fire; an **overdue** reminder fires **once** on catch-up and skips missed windows (matching the trigger tick); a cron that stops computing — or an unparseable `until` — **degrades to a one-shot** rather than running unbounded on a broken value. `until` is inclusive: a slot landing exactly on it still runs.
- Recurring reminders inherit the **once-per-hour floor** from triggers (each wake is a full task activation; i1a08v cost 62). One-shots stay free-form. This reverses a written decision in `docs/plans/v31-triggers.md`, which is now marked superseded there.
- The **30-day cap applies to one-shot `datetime` only** — a recurring schedule is open-ended by design, which is what `until` is for.

## Hardening the fire path this now leans on

A recurring reminder lives indefinitely, so bugs that were survivable for a one-shot stop being so:

- **Re-arm on the live `Task` instance** — fixes a bug that exists on `main`. The path re-armed a detached `loadMetadata()` copy, but `Task.get` returns the live instance for an active task, whose next `debouncedSave` rewrites `metadata.reminder` from stale memory, resurrecting the reminder just fired with a past `trigger_at` that re-fires on restart.
- **Snapshot-iterated tick** — the retry path re-inserts a still-due entry, and a JS Map iterator revisits keys re-added during iteration, so a transient failure spun forever inside one tick.
- **Transient vs permanent `Task.get` failure** — a torn metadata read, a failed migration write or a bad plugins commit used to end a recurrence for the process lifetime while leaving it armed on disk. Now distinguished via `taskExistsOnDisk` and retried next tick.
- **`reminder:fired` is emitted before the re-armed `reminder:set`** — consumers read `fired` as "nothing pending" (`TaskDetail.tsx`), so the reverse order left the ⏰ indicator blank for the whole interval on an armed reminder.
- **The recurrence is spelled out in the wake context** — a fresh subprocess spawns per wake, so without it a cron reminder was indistinguishable from a one-shot and a plain `datetime` call silently destroyed the schedule. A replacing call now says what it replaced, worded per outcome (a one-shot *ends* the repeat; a new cron *supersedes* it).
- **An off switch that doesn't depend on the agent** — `until` bounds the schedule, an unbounded one is flagged in the reply, and `DELETE /api/tasks/:id/reminder` lets an operator cancel directly (404 for an unknown task, idempotent, flushed).

## Known, accepted

Nothing in the task lifecycle clears `metadata.reminder`, and `complete()` is *park*, not terminate — so an unbounded recurring reminder keeps waking a finished task. Clearing on `complete()` would break one-shots (waking a parked task is the feature). `until` + the reply warning + the operator route are the compensating controls; the asymmetry against triggers (no approval, no caps, no kill switch, no listing) is documented in `docs/architecture/triggers.md`.

## Verification

Typecheck clean. Suite **1012/1012** across 80 files.

| File | Tests |
|---|---|
| `src/system/__tests__/reminder-fire.test.ts` | 16 |
| `src/agents/__tests__/reminder-tools.test.ts` | 22 |
| `src/system/__tests__/reminder-rearm.test.ts` | 11 |
| `src/cli/components/TaskDetail.reminder.test.tsx` | 5 |
| `src/connectors/__tests__/trigger-operator-surface.test.ts` | 11 |

Nine mutations across the two rounds — reverted event order, dropped `until` bound, dropped transient retry, live-map iteration, floor note everywhere, dropped replacement warning, the shared contradictory note, dropped 404 guard, dropped `reminder:cancelled` emit — each fail at least one test.

**Live E2E ran this time** (the earlier attempt was blocked by a concurrent instance holding the port). On a booted instance a `*/2` recurrence fired repeatedly, and after cancelling it the next slot passed with no further fire — polled 4.5 min past the slot, fire count stayed put. The no-drift claim also has a probe: 40 consecutive artificially-late fires of `0 5,17 * * *` produced exactly one distinct interval (`[43200000]`), every slot on `T05:00`/`T17:00`.

Two QA items are open by nature, not defects: the mutation-kill claim can't be reproduced by a black-box pass (it requires editing implementation source), and the merge-ordering AC is deploy-only — #244 then #245 then this.


egorkhmelev:
## Added: `cancel_reminder` verified as the off switch, plus a re-arm durability fix (8fda23b)

**Can the agent stop a recurring reminder?** Yes — `cancel_reminder` clears `metadata.reminder` wholesale, so the cron and the next wake both go. Nothing else needs adding. But verifying it exposed a real bug in the fire path.

### The bug

`checkDueReminders` re-armed a `loadMetadata()` copy and wrote it straight to disk, *then* called `Task.get`. But `Task.get` returns the **live instance** for an active task (`activeTasks.get(taskId)`), and that instance owns the whole metadata object — so its next `debouncedSave` (which fires on almost every agent event) rewrites `metadata.reminder` from stale memory, **undoing the re-arm and resurrecting the reminder that just fired**. Its `trigger_at` is in the past, so `rebuildFromDisk` fires it immediately on the next restart.

Narrow for one-shots — a 12h reminder almost always lands on a parked task. Not narrow for recurring reminders, which live indefinitely at up to hourly cadence, so a fire landing on a still-active task stops being an edge case. This PR is what makes it matter.

### Fix

Resolve the task first, re-arm on **that** instance, `save(true)` before reactivating. One authority for the value. The agent then wakes seeing its *next* occurrence — which is exactly what `cancel_reminder` acts on to end the recurrence for good.

### Tests

`src/system/__tests__/reminder-fire.test.ts` — 8 tests: re-arm lands on the live instance, flush-before-reactivate ordering, index stays armed, one-shot still clears, missing task drops cleanly, and two covering `cancel_reminder` as a complete off switch (including cancelling on the wake the reminder just re-armed).

Suite 948/948 across the rebased stack.

PR #260, opened 2026-08-07
Gate Tramline release-lifecycle writes on per-call human approval

fjqtt:
## What this does

Approve/Deny buttons in Slack for Tramline release actions — the same interactive-button mechanism the merge gate uses for PRs. The agent (release-manager) goes to call a Tramline MCP method; if it is a **mutation**, the call is intercepted, a Slack prompt describing the exact action and its target is posted, and the task parks. Someone clicks **Approve** → the agent re-issues that same call and it passes, once. **Deny** → it never runs.

**One rule, no tiers: 9 reads pass, all 36 mutations need a per-call approval.** Nothing is exempt, and nothing is refused-with-approval — the judgement about heavy moves belongs to the approver, whose button states the consequence ("releases to 100% of users immediately — irreversible") and the target's live state.

## How

A `PreToolUse` hook (`src/agents/tramline-guard.ts`) — the same interception layer as the existing filesystem guard, which is what already enforces read-only mode. Approval mechanics copy the merge gate's shape (post buttons → park → resolve → reactivate), with the properties that make a button meaningful:

- **Bound to the exact call.** Approval is keyed on `sha256(action + canonicalized arguments)`: single-use, deduped, 10-min TTL, durable spend. Approving "retry submission A" cannot be spent on B, on another tool, or twice.
- **The prompt is rendered by the engine, not the agent.** Consequence text comes from a per-action map (which doubles as an allowlist: an action with no registered consequence sentence is refused, so a future Tramline tool defaults to *refused*, not to a bare-name button). The target label is built from real `GET /api/v2/releases/:id` payloads via a `PostToolUse` index — `253.0.0 IOS · CI workflow run · release candidate #4821 · (failed)` — and the fixture in the tests is generated from Tramline's serializers, not hand-written. A call whose target this task has not read is refused ("call `get_release` first"). Argument values are sanitized so a string argument cannot append its own lines to the prompt.
- **Same trust model as the other gates.** Any workspace member may resolve the prompt, exactly like edit mode and merge; the clicker's identity is resolved and recorded for the audit trail (external/guest identities omitted, mirroring the merge gate). What keeps the button meaningful is what it says and the single-use digest binding, not who may press it.
- **Fail closed.** Any error while evaluating a gated call — including agent-controlled input that overflows the canonicalizer — is a deny, not a throw.
- **Slack is the only approver surface.** `POST /tasks/:id/approve` rejects `type: 'tramline_action'`: the `/api` router is unauthenticated and its SSE stream publishes the digest, so an HTTP path would bypass the allowlist.

Everything else is untouched: the hook returns `continue` for any non-`mcp__tramline__*` tool, all changes to `task.ts`/Slack handlers/types are additive, and with a read-only Tramline key every approved call still 403s server-side.

## Review history

Four adversarial reviews ran against earlier revisions of this branch; every confirmed finding is fixed in the current diff. The two that shaped it most: the label index was originally written from Tramline's domain model instead of its serializers (iOS/Android prompts rendered identically — now fixture-tested from real payloads), and two attempts at classifying "safe" vs "forbidden" writes both collapsed against Tramline's actual code (the "safe" tier had members that start production releases; the "forbidden" list wasn't closed under effect). The uniform rule is the operator's call and also the only rule that cannot drift.

## Config

| Variable | Note |
| --- | --- |
| `MCP_TRAMLINE_API_KEY` | must carry Tramline's **`write`** scope for approved calls to succeed (currently read-only) — and swapping back to a read-only key is the rollback switch: every approved call then 403s server-side. |

## Merge & deploy order — load-bearing

Plugins hot-reload on every task load (`syncPlugins` does `git reset --hard origin/main`); this engine ships in a deployed image. So: **(0) merge and deploy `sweatco/tramline#139` to the Tramline instance first** — without it `end_soak`/`extend_soak` 422 on the API path and approved actions stamp as "Tramline" instead of the key's name — **then deploy this → verify live that a gated call produces a button and no HTTP reaches Tramline → then merge `archie-plugins#140` → then, and only then, switch the key to `write`.** The one assumption with no automated check yet is that the CLI delivers `mcp__*` calls to `PreToolUse` hooks with the expected shapes — same class of version-coupled control as the egress allowlist (`tools/e2e/egress-check.ts`); an equivalent live check for this gate is the top follow-up.

## Testing

`tsc` clean; full suite 74 files / 1031 tests green, 58 of them in `tramline-guard.test.ts` (classification of all 36 mutations, digest binding, label uniqueness on the serializer-derived fixture, prompt-forging resistance, fail-closed paths, allowlist fail-closed, TTL/dedup/stale-click behaviour).

## Related

- `sweatco/tramline#139` — API-key attribution on the release timeline + `[Mutation]` markers in the MCP tool descriptions
- `sweatco/archie-plugins#140` — removes the agent-side Tramline denylist (the gate replaces it), updates agent/PM guidance, vendors the rebuilt bundle


egorkhmelev:
@fjqtt thank you for the improvement, however this is quite specific MCP related change in a generic engine, which should abstract from such things. I don't think we will merge this as project specific stuff either done in a generic fashion or live in project's plugins repo. What you want should be covered by addressing it in a generic fashion like requested in #168 issue. Feel free to take it.

fjqtt:
Agreed — reworked in the generic shape #168 asks for: #263 (draft). The Tramline-specific pieces move where they belong: the tool policy (tiers + button titles) into the release-manager agent's frontmatter in archie-plugins, and the consequence descriptions into the Tramline MCP server's own tool docs (sweatco/tramline#139). This branch stays as the porting source; the review history here (four adversarial rounds) is baked into #263's invariants.

PR #263, opened 2026-08-07
Human-in-the-loop approval gate for critical MCP tools

fjqtt:
Implements #168 — a human-in-the-loop approval gate for critical MCP tools. Draft: the mechanism is complete, tested and documented; what wants a maintainer's eye is the config shape, which this revision deliberately simplified.

## The config

One block per server, in the plugins repo's root `.mcp.json`, next to the connection config it governs:

```json
"tramline": {
  "command": "node",
  "args": ["${MCP_TRAMLINE_SERVER_PATH}"],
  "description": "Tramline — mobile release management",
  "archie": {
    "default": "ask",
    "allow": ["list_apps", "get_release", "get_release_analytics"],
    "deny":  ["start_release", "stop_release"]
  }
}
```

| Tier | Meaning |
|---|---|
| `allow` | runs with no gate |
| `ask` | a human approves **that specific call**, arguments shown, in Slack |
| `deny` | never runs — and is withheld from every agent that mounts the server |

`default` covers everything unlisted and is `ask` when omitted, so a tool the vendor ships next quarter arrives gated rather than silently open. Invert it per server: list the safe reads for a mostly-dangerous server, list the exceptions for a mostly-safe one. The tier names are Claude Code's own permission vocabulary — no new concepts to learn.

`ask` is the whole point of the feature: it is the state that does not exist today. Right now a tool is either allowed and then invoked with no human check, or blocked outright — which is why every tool that mutates an external system is currently just switched off (`release-manager.md` carries 59 hand-listed denials, and a comment reading "re-running a failed job is a planned capability, deliberately not enabled yet").

## Default behaviour does not change

**A server with no `archie` block is unmanaged: no hook is attached, and every one of its tools behaves exactly as before.** Rollout is opt-in per server, no existing plugin changes meaning, and an agent whose servers are all unmanaged runs the same code path it ran before. The fail-safe `ask` default applies only *within* a server someone has already opted in.

## The policy belongs to the server, not the agent

Which of TeamCity's tools are dangerous is a fact about TeamCity. Declaring it once per server covers every agent that mounts it, and three things fall out for free:

- **The PM is covered** like any other agent — its overlay resolves servers through the same `resolveAgentMcpServers`. Under a frontmatter-based policy it silently wasn't, and it is the one agent holding the admin-panel server.
- **`deny` replaces the copy-pasted denylists.** In `archie-plugins`, mobile's 23 `disallowedTools` entries are a *strict subset* of release-manager's 59 — the same facts about the same three servers, maintained twice. There is currently no agent pair that diverges at all.
- **Renaming a server key moves its policy with it.** No second place to update, no orphaned policy.

Frontmatter `disallowedTools` still works and merges on top, for the case where one agent must not touch something its peers may (`restore_streak_days` on growth-ops/ops, which the PM must keep — see the consumer PR).

## What the approver reads

The heading is the policy's optional **title** for the tool, then the call's real arguments:

```
Release this rollout to 100% of users immediately — irreversible
Tool: `tramline:fully_release_rollout`
Arguments: id=`226284f5-…`
```

```json
"archie": {
  "default": "ask",
  "titles": { "fully_release_rollout": "Release this rollout to 100% of users immediately — irreversible" }
}
```

Write one only where the method name is a bad button on its own — `fully_release_rollout` and `fully_release_previous_rollout` are one word apart and mean very different things. An untitled tool renders as `Run server:tool`: terse but honest, with the sanitized arguments either way.

Both halves are rendered and sanitized **by the engine, never by the agent** — an agent's own summary of what it intends is never what the human reads. Argument values are the one part it controls, and unescaped a string argument can append its own lines to the message ("*Note:* pre-agreed, safe to approve"); values and titles alike are flattened, sigil-stripped and capped, arguments are code-spanned, and their count is capped too because Slack refuses a section over 3000 characters. Proven by test. An agent that wants to explain itself does so as its own message in the thread, beside the button rather than inside it.

## Mechanism

- **`PreToolUse`, not `canUseTool`.** Every agent runs under `permissionMode: bypassPermissions`, which the SDK documents as auto-approving calls past `canUseTool` ("PreToolUse hook denies bypass canUseTool", sdk.d.ts). The hook is the same layer the filesystem guard already trusts for read-only mode. Consequence: no pause-and-resume of the original call — deny, then a single-use grant that the agent's retry spends, so the action always runs through the same audited MCP path.
- **Approval is bound to the exact call**: `sha256(server, tool, canonicalized args)`. It cannot be redirected to another tool, other arguments, or spent twice.
- **Lifecycle invariants**, each present because an adversarial review of the predecessor branch broke the version without it: one pending request per task, 1h age-out, no supersede; a failed Slack post clears the slot (else the task wedges refusing everything for an hour); grants dedupe by digest; the spend is durable while grants are debounced (a crash must not resurrect a spent single-use token); an expired click is a stale no-op that mints nothing; findings for requested / approved-by-name / denied / expired-unspent / actually-spent.
- **Fail closed.** Any error evaluating a managed call is a deny. Unmanaged tools are classified before the try block and proceed even when the gate's own dependencies are broken.
- **Strict validation.** A malformed `archie` block fails the load — unknown tier, unknown key, a tool in two tiers. A misspelled tier (`asks:`) would otherwise read as "nothing listed", quietly dropping every tool in it to the default. A server *key* that cannot survive the `mcp__<server>__<tool>` round trip (one containing `__`) is rejected too: the gate would never match its calls, so every tool of that server would run ungated.
- **Trust model = the existing gates.** Any workspace member may resolve via Slack (identity recorded for the audit trail, external/guest identity omitted, exactly like merge and edit mode), and the CLI/API path resolves with `type: 'tool_call'` + `ref` (the digest) like every other approval type — which is also what makes the e2e harness able to drive it.

## Review round

Reviewed by an independent pass focused on stability and minimalism; the findings that survived verification are fixed in this revision:

- **Park-with-no-button (major).** The failure path that clears the pending slot covered only the Slack post, while a durable flush and an awaited findings-write sat between setting the slot and that try block. Either throwing left the slot set, and the same-digest retry then parked the task against a prompt that was never posted. The findings write is now fire-and-forget (as the spent-finding already was) and the catch covers the flush.
- **Unenforceable policy on an awkward server key (major).** A key containing `__` loaded happily and then never matched any call — silently ungating the whole server. Now rejected at load, with a test.
- **The spend was not actually durable.** It was flushed fire-and-forget, so a crash after the tool ran could leave the single-use grant spendable again — which the docs claimed was closed. The splice stays synchronous (that is what stops a double spend in one turn); the write is now awaited before the call proceeds, and a failed write logs and proceeds rather than refusing an approved action.
- **A cross-agent retry parked an agent nothing would unpark.** Both resolution paths clear the requester's teardown alone, so a second agent hitting the same live slot armed a deferred stop that fired mid-retry. It is told to wait instead.
- **Hardening:** own-property lookups in `classifyToolCall` (a tool named `constructor` reached a prototype value outside the fail-closed try); the rendered argument list is capped by count as well as by value length (an uncapped list overflows Slack's 3000-char section limit and makes the call permanently unapprovable); `null` arguments are now shown, since the digest covers them and hiding one had the approver approve an argument they never saw.
- **Deleted:** an `ApprovedToolCall.heading` field that was written and never read, a `proceed()` helper whose comment justified it with something untrue, the hand-rolled copy of `mcpToolName` in `spawn.ts`, ~30 lines of module-header prose that restated the architecture doc, and 4 redundant test cases.

Left deliberately: a malformed policy in the plugins repo throws inside every `Task.get`, not only at boot (fail-closed and loud beats a half-applied policy, but validating at sync time and keeping the previous config would fail equally closed — worth deciding separately); and the two near-clone Slack button handlers, which mirror the existing merge-handler shape.

## Changes since the first revision

The mechanism is unchanged; the configuration surface shrank by about two thirds.

- **Policy moved from agent frontmatter (`metadata.archie.mcpPolicy`) to the server entry in `.mcp.json`** — the reasons are in the section above. Verified against Claude Code 2.1.237 that an unknown key inside a server entry is ignored with no diagnostics, while a genuinely invalid entry is reported and skipped, so the file stays usable by Claude Code itself. `description` had already established the precedent; both keys are stripped before the SDK sees the config.
- **`requireTitles` is gone; `titles` is now optional.** The first revision required a title for every gated tool. The second replaced titles entirely with the tool's own description reported by its MCP server — which the live run then proved unreachable (see below), so titles are back as an optional per-tool override and the untitled case falls back to the bare `server:tool` identity.
- **The `rw` tier is gone.** It had no consumer, and "silently denied unless edit mode happens to be on" is a confusing thing for an agent to hit. It can come back if a case appears.
- **Tiers renamed** `auto`/`human` → `allow`/`ask`, plus `deny`, matching Claude Code's permission vocabulary.
- **The PM is now covered**, as a consequence of the above rather than as a fix.
- **Fixed: the gate source contained two literal NUL bytes**, which made git treat `src/agents/tool-approval-gate.ts` as binary — GitHub showed `0 additions, 0 deletions` for the single most security-relevant file in the earlier revision. The digest now joins its parts as a JSON array; the file is plain text and reviewable.
- **Rebased onto current main** (the first revision was 136 commits behind) and squashed to one commit.

## Live verification

Booted from this branch and run against a real instance (attested `git_sha` matching the checkout), because the one thing no unit test can see is whether the Claude CLI still delivers MCP tool calls to `PreToolUse` under `bypassPermissions`:

```
tool-gate-check: approval requested (ref de72f056d7263e25); marker count now 0
tool-gate-check: approved — waiting for the retry to spend the grant
tool-gate-check: PASS
  - gated MCP call intercepted by PreToolUse (approval requested, nothing executed)
  - approval spent exactly once; readonly tool passed ungated
```

The audit trail from that task, and the marker file — bind-mounted, so non-execution is observable from the host — holding exactly one line:

```
[decision]   Tool-call approval requested: gatecheck:write_marker — Append a marker line to the shared e2e marker file
[decision]   Tool call approved: gatecheck:write_marker — Append a marker line to the shared e2e marker file
[completion] Gated tool call ran on approval de72f056d7263e25: gatecheck:write_marker
```

**What that run caught.** On the first pass the heading read `Run \`gatecheck:write_marker\`` — the fallback — even though the stub server returns a description for that tool, which I had confirmed by talking to it over stdio. A minimal repro inside the container explains it: `mcpServerStatus()` reports the server `connected` with both tools, but `description` is absent on every entry and `annotations` is `{}`. The CLI has the descriptions (it gives them to the model) and exposes them to the SDK host nowhere — that field is the only one in the whole SDK type surface that would carry them, and it comes back empty. So the description-based heading was dead code in practice, and the mechanism I had used to justify deleting titles does not exist. Titles are back, optional, and the `docs` say why with the date. Side finding, pre-existing and not from this branch: the same empty `annotations` means `readOnly` — used elsewhere to phrase the Slack status line — is getting nothing either.

## Testing

`tsc` clean; full suite 96 files / 1528 tests green. New coverage: the gate (tiers, digest binding, rendering/sanitization of both arguments and descriptions, fail-closed paths), the task lifecycle on the real `Task` prototype methods via the merge-approval harness pattern, `.mcp.json` policy parsing and its rejection cases, and the registry wiring (policy attaches to every mounting agent, does not leak to servers the agent has not mounted, `deny` lands in `disallowedTools` deduped, and the PM is covered through its overlay).

Plus a **live tripwire**: `tools/e2e/tool-gate-check.ts` with the `gatecheck` example plugin (stub MCP server, dependency-free) drives a booted instance over the HTTP API and asserts interception, deny-blocks-execution (a marker file under the bind-mounted workdir proves non-execution from the host), single-use spend, and that the `allow`-tier tool passes ungated. Same rationale as the egress check: `PreToolUse` delivery for MCP tools lives in the Claude CLI, is version-coupled, and silently regressed across a CLI bump once. Documented as section 3c of the `archie-e2e` skill. Run it after any SDK bump before trusting the gate with a write-scoped credential.

## Consumer

`sweatco/archie-plugins#140` is the pilot, rewritten onto this schema. Tramline's entire rule becomes twelve lines in `.mcp.json` — `default: ask` with its nine reads under `allow` — so its 36 write tools move from "flatly denied" to "held for per-call approval", which is the capability that PR has been waiting on. Nothing outside Tramline changes: the other 12 agents' effective permissions are byte-identical, and every other server keeps its denials in agent frontmatter untouched.

That PR also carries the vendored `tramline-mcp` bundle with honest per-tool descriptions from `sweatco/tramline#139` — which matters more under this revision than it did before, since the approval button's heading *is* the tool's own description now. Merge order is load-bearing and documented there: tramline#139 deployed to the instance (it merged on 7 Aug) → engine deploy → verify a live button → plugins → swap `MCP_TRAMLINE_API_KEY` to write scope last.

## Relation to #260

#260 was the Tramline-specific first cut, which went through four adversarial review rounds; closing it in favour of this. Every property listed above is regression-tested here.

## Open questions

1. **Config shape** — `default` + `allow`/`ask`/`deny` per server in `.mcp.json`. Comfortable? The alternative considered and rejected was a sidecar `mcp-policy.json` keyed by server name: zero risk to the `.mcp.json` schema, but two files to keep in sync and mandatory key-drift validation.
2. **Per-agent divergence has no expression** beyond `disallowedTools`. No current agent pair needs one; the coarse tool meanwhile is a second server entry with its own credential (`tramline` read-only, `tramline-rw` gated), which also gets credential scoping right. Confirm it stays out of v1?
3. **Reusable approvals** ("approve this tool for the rest of the task") — listed in the issue as a possible follow-up. Not included; the per-call default is the point. Confirm it stays out?
4. **A grant is not bound to the requesting agent** — the digest covers server, tool and arguments, so another agent on the same task making the byte-identical call could spend it. Reading it as "the *action* was approved" is intentional, but `requested_by` in the audit trail then names the requester rather than the executor. Tighten, or document as is (currently documented)?
5. **Redaction** of secret-looking argument values in the preview (the issue's open question) — not addressed beyond sanitization and capping. v1 or follow-up?

### DIFF
--- a/.claude/skills/archie-e2e/SKILL.md
+++ b/.claude/skills/archie-e2e/SKILL.md
@@ -161,6 +161,16 @@ Why it exists: the allowlist is enforced by the Claude CLI, not by our code, and
 
 A failure here is a release blocker, not a flake. If the non-allowlisted host is reachable, agent egress is open.
 
+## 3c. Tool-approval-gate check (run on any branch that touches the gate, spawn hooks, or bumps the SDK)
+
+```bash
+npx tsx tools/e2e/tool-gate-check.ts     # requires a booted instance
+```
+
+Asserts the MCP tool approval gate end to end on the live instance, using the `gatecheck` example plugin (a stub MCP server whose one mutation appends to `workdir/e2e/gate-marker.log`, observable from the host; its policy, including the button title, is the `archie` block on the server in `examples/plugins/.mcp.json`): a gated call must produce a `tool_call` approval request **without executing**, approving via the API must let the agent's retry execute **exactly once**, and the fixture's `allow`-tier tool must pass ungated. Same rationale as the egress check: the interception lives in the Claude CLI (`PreToolUse` under `bypassPermissions`), is version-coupled, and no unit test can notice the CLI silently ceasing to deliver MCP calls to hooks. A failure is a release blocker for anything relying on the gate.
+
+Prerequisite: the booted workdir includes the `gatecheck` example plugin (`npm run example:setup` links `examples/plugins`, which ships it). The check is self-driving over the HTTP API — no MCP session needed — and takes one model round-trip per phase (~2-5 min total).
+
 ## 4. Teardown
 
 ```bash
--- a/docs/architecture/edit-mode.md
+++ b/docs/architecture/edit-mode.md
@@ -2,7 +2,7 @@
 
 Archie operates in two modes: **readonly** (the default) and **edit** (after human approval). This two-mode system implements a human-in-the-loop safety gate that prevents agents from modifying code without explicit user consent.
 
-> A sibling gate, [max mode](max-mode.md), reuses this same request → approve → respawn shape to upgrade the coding agents' model/effort for a task. The two are independent — a task can have either, both, or neither.
+> A sibling gate, [max mode](max-mode.md), reuses this same request → approve → respawn shape to upgrade the coding agents' model/effort for a task. A second sibling, [tool approvals](tool-approvals.md), gates individual MCP tool calls on per-call approval rather than a task-lifetime grant. All are independent.
 
 ## Two-Mode System
 
--- a/docs/architecture/overview.md
+++ b/docs/architecture/overview.md
@@ -243,6 +243,7 @@ prompts/                         # Repo-root: layered system prompts
 - [GitHub Integration](github-integration.md) -- webhooks, PR management, merge orchestrator
 - [Edit Mode](edit-mode.md) -- approval flow, shared clones, and git workflow
 - [Max Mode](max-mode.md) -- per-task, human-approved model/effort upgrade for coding agents
+- [Tool Approvals](tool-approvals.md) -- per-call human approval for critical MCP tools
 - [Plugin System](plugin-system.md) -- plugin structure, loading, and agent registration
 - [Web Research](web-research.md) -- multi-agent research pipeline and defense layers
 - [Security](security.md) -- research budget, sandwich defense, prompt injection mitigations
--- a/docs/architecture/plugin-system.md
+++ b/docs/architecture/plugin-system.md
@@ -103,6 +103,11 @@ A plugin named `pm` is treated specially. Its `agents/pm.md` file provides:
 
 A single `.mcp.json` file at the plugins directory root (`$ARCHIE_WORKDIR/plugins/.mcp.json`) provides MCP server connection configs. Individual agents reference server names from this file via their frontmatter `mcpServers: [...]` field. Environment variables matching `${MCP_*}` are substituted at load time.
 
+A server entry may carry two Archie extensions, both parsed by `loadMcpJson` and **stripped before the config reaches the SDK** so a plugin authored for Archie stays a valid Claude plugin:
+
+- `description` — one human-readable line, surfaced in the PM's roster of what each teammate can reach.
+- `archie` — the tool approval policy for that server: `{ default, allow, ask, deny }`, each tier a list of bare tool names, plus an optional `titles` map giving the approver-facing button text for tools whose name is a bad button on its own. Because it lives with the server, every agent that mounts the server inherits it (the PM included), and `deny`-tier tools are appended to that agent's `disallowedTools`. A server without this block is unmanaged and behaves exactly as before the gate existed. A malformed block fails the load. See [Tool Approvals](tool-approvals.md).
+
 ## Plugin Loader
 
 The plugin loader (`src/system/plugin-loader.ts`) runs once at startup using synchronous filesystem reads. It scans every subdirectory of `PLUGINS_DIR` and produces a `LoadedPlugin[]` array consumed by downstream modules.
--- a/docs/architecture/security.md
+++ b/docs/architecture/security.md
@@ -201,6 +201,12 @@ In edit mode, repo agents manage their own PRs directly via the `repo-tools` MCP
 
 **Source:** `src/agents/spawn.ts`, `src/agents/tools.ts` (`createRequestEditModeTool`)
 
+### MCP Tool Approval Gate
+
+An MCP server can declare, per tool, that calls need per-call human approval — an `archie` block next to its connection config in the plugins repo's `.mcp.json` (see [Tool Approvals](tool-approvals.md)). A `PreToolUse` hook classifies each call against the policy of the servers that agent mounts: `allow` runs ungated, `deny` never runs (and is withheld through `disallowedTools` up front), `ask` is denied while a Slack Approve/Deny prompt — rendered by the engine from the policy's optional per-tool title and the call's sanitized arguments, never from the agent's summary — is posted and the task parks. Approval stores a single-use grant bound to `sha256(server, tool, canonicalized args)`; the agent's retry of that exact call spends it once. A server with no `archie` block is unmanaged and behaves exactly as before the gate existed. Because the policy travels with the server, every agent that mounts it is covered by the same rules, the PM included. Resolution follows the same trust model as the other gates (any workspace member; identity recorded for audit); the `/api` path is unauthenticated, as with every other approval type.
+
+**Source:** `src/agents/tool-approval-gate.ts`, `src/tasks/task.ts`, `src/connectors/slack/events.ts` (`registerToolApprovalHandlers`)
+
 ### PR Review Enforcement
 
 All code changes go through GitHub pull requests. Repo agents create PRs via the `create_pull_request` tool, and merge is gated on external PR review approval. The `merge_pull_request` tool checks that PRs are approved, CI is passing, and there are no conflicts before merging.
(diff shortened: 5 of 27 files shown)


fjqtt:
**Live verification: the gate holds on a real instance.** Ran the new tripwire against an instance booted from this branch (`tools/e2e/boot.ts`, checkout attested: `git_sha 9d2952d`):

```
tool-gate-check: instance http://localhost:3000, nonce E2E-GATE-f1e71984
tool-gate-check: task task-20260817-1140-if7yyc
tool-gate-check: approval requested (ref 724d9ecaea45a518); marker count now 0
tool-gate-check: approved — waiting for the retry to spend the grant
tool-gate-check: PASS
  - gated MCP call intercepted by PreToolUse (approval requested, nothing executed)
  - approval spent exactly once; readonly tool passed ungated
```

Container-side trace of the same run:

```
[gatekeeper-agent:ro] gatecheck: get_status                                  ← readonly, passed ungated
[gatekeeper-agent:ro] gatecheck: write_marker value=E2E-GATE-f1e71984        ← gated call attempted
[Slack] POST (interactive): Approve tool call: Write the e2e marker '`E2E-GATE-f1e71984`' …
```

Marker file after the run: exactly one line, `E2E-GATE-f1e71984`.

This settles the one assumption no unit test could reach: the Claude CLI (bundled 2.1.x) **does** deliver `mcp__*` calls to `PreToolUse` hooks under `bypassPermissions`, `deny` **does** block the call before it reaches the MCP server (marker count was 0 at approval time), and the digest-bound grant is spent exactly once by the agent's retry. Teardown clean.

egorkhmelev (approved):

Issue #265, opened 2026-08-07
Canvas edited mid-task is not re-shown to a task already briefed on that channel


egorkhmelev:
Follow-up from #258, which is a deliberate trade rather than a defect — filing it so the decision is revisitable.

## Behaviour

`post_to_channel`'s preflight shows a destination channel's standing brief once per task, tracked in `TaskMetadata.briefed_channels`. Once a channel is on that list, later posts to it from the same task skip the preflight entirely — including the `ensureChannelCanvas` scan.

So if someone edits the destination channel's `Archie…` canvas *while a task is in flight*, a task that has already posted there keeps working from the brief it read the first time. The edit reaches new tasks immediately; it never reaches that one.

## Why it is this way

Originally the set lived on the `Agent`, which re-showed the brief on every re-activation — a settled task leaves the active registry, and the next message rebuilds `Task` from disk with a fresh Agent. Live testing showed the same brief being handed to the same thread twice, the second time to a message that already complied with it. Moving to task metadata fixed that, at the cost of never re-reading within a task.

## Options

1. **Leave it.** Canvases are standing context and rarely change mid-task; new tasks pick edits up within the 60s canvas TTL.
2. **Re-show when the canvas changed.** Store `updatedTs` alongside the channel id in `briefed_channels` and re-brief when `files.info.updated` advances. Costs one `files.info` per cross-channel post — the same call the preflight already makes on a first post — so cheap, but it reintroduces a round-trip whenever an edit lands mid-task.
3. **Re-show on a TTL.** Time-box the suppression. Simplest to reason about, but re-briefs on a timer rather than on an actual change.

Option 2 is the closest to intent: suppress the *repeat*, not the *update*.

## Not urgent

The failure mode is a stale brief within one task's lifetime, not a wrong or unsafe post — the trust rules (destination brief narrows, never widens) are in the note itself and were verified live on #258, independent of freshness.

Issue #241, opened 2026-07-28
Give the PM the preceding channel messages as ambient context when a task starts from a top-level mention


egorkhmelev:
### What problem does this solve?

People routinely type into the channel instead of into a thread: two or three top-level messages describing something, then a final message that `@`-mentions Archie. Those earlier messages are separate top-level posts, not thread replies, so they are invisible to the task that gets created.

Mechanically: `handleSlackEvent` collapses everything into `threadId = event.thread_ts || event.ts` (`src/connectors/slack/events.ts:662`), and `fetchSlackThread` reads `conversations.replies(ts=threadId)` (`src/connectors/slack/client.ts:1558` → `:566`). When the triggering message *is* the thread starter (no `thread_ts`), that call returns exactly one message — the mention itself. The PM spawns with a single line of context and either asks "what are you referring to?" or guesses.

Thread replies are already fine: the full thread history lands in `knowledge.log` via `Task.append` (`src/tasks/task.ts:353-370`). The gap is only the thread-starter case.

### Proposed solution

When a task is created from a message that is a thread starter, snapshot the last ~3 channel messages posted *before* it and give them to the PM as clearly labeled ambient context — explicitly not part of the task.

1. **Detect** in `handleSlackEvent`, on the new-task branch only, that the trigger is the root (`!event.thread_ts`). No change for thread replies or for messages routed to an existing task.
2. **Fetch** the preceding messages with the existing wrapper `fetchChannelHistory` (`client.ts:1085`). `channels:history` / `groups:history` are already granted, so no manifest change and no reinstall. The wrapper has no `latest` param today — either add one or over-fetch and drop everything with `ts >= event.ts`.
3. **Bound it**: N ≈ 3 messages *and* a recency window (~60 min). Chatter from yesterday is noise, not context. Reuse the ingestion filters — drop own-bot messages, drop external bots, redact external authors in shared channels (`renderMessageForContext`, `src/tasks/persistence.ts:242-247`).
4. **Persist** the snapshot on the task (in metadata, alongside the channel entry) so it survives respawn/recovery, and render it at spawn as its own prompt block. Do **not** append it to `knowledge.log` — that file is the task's own record and drives `last_processed_ts`; ambient chatter does not belong in it.
5. **Frame** it the way the canvas injection already does (`buildChannelCanvasPromptSection`, `src/connectors/slack/channel-canvas.ts:167`) — an XML block whose `note` attribute carries the epistemic status:

```
[...]

PR #269, opened 2026-08-09
feat(slack): channel pinned messages as agent context

egorkhmelev:
## What & why

Slack pins are a standing signal about a channel in the same way the `Archie…` canvas is, but a much noisier one: nobody writes a pin for Archie, they accumulate for years, and anyone in the channel can add one. Dumping their contents into the prompt would be expensive and mostly stale; ignoring them loses the runbook a team pinned two years ago and still lives by.

So this injects a one-line **index** of what a channel pinned — with both dates and both ages on every line — and lets the PM open any individual pin when the index says it matters. Recency is exposed, never filtered: an old pin is often the most relevant thing in a channel, and only the agent can judge that.

Groundwork for #241 in shape, though it solves a different problem.

## How it works

`ensureChannelPins` runs off the same two `events.ts` hooks as the canvas scan, TTL-bounded to once a minute per channel, and posts nothing — indexing a pin changes nothing for the channel, so there is no state change worth announcing (PR #254). `buildChannelPinsPromptSection` renders the block at a single injection point in `spawn.ts`, so PM, repo and plugin agents all get it.

Four decisions worth following the diff for:

**Opening a pin needs no new tool.** `read_thread` already returns a reply-less message as itself, and `fetch_slack_reference` now accepts a pinned file id alongside a canvas-referenced one. Both are PM-only, so a specialist that needs a pin asks — exactly as it does for a canvas file reference.

**Two principals are gated, not one.** A pin has an author and a pinner, usually different people; either being external drops it, either being unclassifiable fails closed. Bot-posted pins are adopted and marked `by="<name> (app)"`, with only bots from *another* workspace refused — the same line thread ingestion already draws at `client.ts:2200`. The pinner is what makes that safe: a bot has no human author to vet, but a person always chose to put the line in front of the agent.

**Gate first, cap second.** Capping the 25 newest before gating let refused pins consume index slots, so one external participant pinning twenty-five things blanked a channel's index entirely. The omission count is computed post-gate, so the block discloses what the *cap* hid and never what the gate refused.

**Containment is XML escaping.** The first two attempts stripped closing tags out of the text; adversarial QA broke both. Only closing tags were neutralised, so a body could forge a `<pin by="someone else">` element; `JSON.stringify` escapes a quote as a JSON escape, which means nothing in tag syntax; and the invisible-character blocklist caught 8 of 197 candidate codepoints. Every value now goes through one escape covering `&`, `<`, `>` and `"`, which holds without enumerating attacks.

Summaries are cheap: a pin at or under 200 characters is its own index line with no model call, a longer one gets a single Haiku call cached against a digest of its text, and at most six run per scan because the scan is awaited before the PM wakes. Each line carries `source` — `verbatim` means it is the pinned text itself, reaching the prompt unaltered, which the block's note tells the agent to treat as untrusted input.

**Out of scope, deliberately:** no `pin_added`/`pin_removed` subscription (TTL scan only), no age filtering or ranking beyond pin recency, no writing pins, and #265 is inherited rather than fixed — a pin added mid-task appears at the agent's next wake.

## Verification

`npm run typecheck`, `npm run build` and `npm test` are all clean: **78 files, 1108 tests**, up from 1037 on `main`.

Every criterion was verified by agents that had not seen the diff, and the evidence was audited by another that ran nothing. The manifest below is computed, not written.

**This was then run live against the real Sweatcoin workspace**, after `pins:read` was granted — see the section below the manifest. AC18 shipped as a waiver and is now verified; four defects that every unit and structural check had passed came out of that run and are fixed on this branch.

| AC | Criterion | Method | Status | Where it ran |
|---|---|---|---|---|
| AC1 | WHEN a channel has pinned items and the bot holds pins:read, THEN every agent spawned for a task linked to that channel receives a <channel_pinned_messages> block listing them, newest-pinned first. | unit | **verified** | command |
| AC2 | WHEN pins.list fails for any reason including missing_scope, THEN no pins block is emitted for a channel with nothing stored, any previously stored pins are left untouched, the task spawns normally, and nothing throws. | unit | **verified** | command |
| AC3 | WHEN a pinned item's author OR its pinner is external/guest, THEN that item is excluded from the block and from the fetch allowlist. | unit | **verified** | suite |
| AC4 | WHEN either principal's classification cannot be determined, THEN the pin is not newly adopted, and a previously adopted entry is retained unchanged. | unit | **verified** | suite |
| AC5 | WHEN a pin's normalised text is at most 200 characters THEN its index line carries that text verbatim; WHEN longer THEN a one-line Haiku summary is used, and a Haiku failure falls back to truncation rather than dropping the pin. | unit | **verified** | command |
| AC6 | WHEN a pin has been summarised, THEN the summary is persisted in that pin's channel-store entry together with its key and a digest of the source text, and it survives a process restart — a reloaded store produces no new model call for that pin. | unit | **verified** | suite |
| AC7 | Every index line carries the pin date and age, the message date and age, the author, the pinner, and the identifier needed to load it — ts for a message, file id for a file. | unit | **verified** | command |
| AC8 | WHEN a channel has more than 25 pins, THEN the 25 most-recently-pinned are shown and the block states how many were omitted for that channel. | unit | **verified** | command |
| AC9 | WHEN a pinned item is a file, THEN it appears in the index and its file id is accepted by fetch_slack_reference; a file id neither pinned nor canvas-referenced is still refused. | unit | **verified** | suite |
| AC10 | Pinned content cannot escape its wrapper: a pin whose text contains </channel_pinned_messages> or </pin> is neutralised. | unit | **verified** | command |
| AC11 | WHEN a channel store written before this change is loaded (no pins field), THEN it loads without error and behaves as 'no pins yet'. | unit | **verified** | suite |
| AC12 | Scanning pins posts nothing into the channel — no adopt/ignore/drop announcement, no reaction, no ephemeral. | unit | **verified** | suite |
| AC13 | The pins scan is TTL-bounded: a second inbound event inside the window makes no further pins.list call. | unit | **verified** | suite |
| AC14 | slack-manifest.yaml declares the pins:read bot scope. | structural | **verified** | suite |
| AC15 | The pins block has exactly one injection point, reaching PM, repo and plugin agents alike. | structural | **verified** | command |
| AC16 | prompts/pm-agent.md, prompts/agent-core.md and skills/channel-canvas/SKILL.md state what the block is, that it may be stale, and that only the PM can open a pin. | structural | **verified** | command |
| AC17 | npm run typecheck, npm run build and npm test all pass. | structural | **verified** | command |
| AC18 | Against the real workspace after manifest re-import and reinstall, a pinned message appears in the PM's context and read_thread on its ts returns the full message. | deploy-only | **waived** | none |
| AC19 | WHEN a pinned message's text changes, THEN the digest mismatch triggers exactly one re-summarisation, and the new summary replaces the old one in the store. | unit | **verified** | suite |
| AC20 | WHEN an item is unpinned in Slack, THEN it disappears from the stored pins, from the rendered block and from the fetch allowlist on the next scan — while a failed pins.list still leaves every stored pin intact. | unit | **verified** | suite |

### Live verification

> No criterion required a live instance, so nothing was expected to run against one.

To be plain about why: the E2E harness cannot reach this feature at all. Tasks created through the debug MCP are CLI-channel-only, so every channel-context builder returns `''` by construction. That was established during discovery and accepted at sign-off, not discovered late.

### Live run against the real workspace

After `pins:read` was granted and the app reinstalled, this branch was booted from the worktree and driven through real Slack events in `#bot-test`. Nine checks, all passing:

| # | Check | Evidence |
|---|---|---|
| 1–4 | Boots on this branch, task flow unaffected, no crash, no spurious block | SHA-attested boot; task created and answered |
| 5 | Index appears, newest-pinned first, both dates and both ages | 3 pins indexed from a real channel |
| 6 | `source` honest | 25-char pin → `verbatim`, byte-identical; 2509-char brief → `model`, 87 chars |
| 7 | PM opens a pin | Its reply quoted regions, step goal, prize and dates that appear only in the source message, not the index |
| 8 | TTL and digest reuse | Follow-up 30s later left `pinsCheckedAt` unmoved — no `pins.list` at all; a genuine later scan returned byte-identical summaries and digests |
| 9 | Unpin | `pinsEligible` 3 → 2, item gone from store, block and allowlist |

On (7), the PM's own words, unprompted: *"The index itself is just a pointer list with dates and authors — I don't treat a line as instruction; I open the message before acting on anything in it."* That is the block's `note` doing its job.

**Four defects the live run found that unit and structural checks had all passed.** Each is fixed and covered on this branch:

- **Bot-posted pins were refused outright.** The first real scan refused an Archie post a human had pinned. The rule was aimed at relay apps laundering outside content, but it gutted the feature for its highest-signal input and diverged from `client.ts:2200`. Now only foreign-workspace bots are refused.
- **A summary came back as JSON.** A pin whose body was JSON produced the index line `{"entry": "Daily recap covering …"}` — the model mirrored its input's shape into the string field it was asked to fill. Schema validation cannot catch that; a JSON blob is a valid string. Unwrapped now, in the unambiguous case only.
- **`<1d` was read as "yesterday".** Three pins made 22 minutes earlier rendered `pinned_age="<1d"`, and the PM told the person who had just pinned them that they were pinned yesterday. Ages now carry sub-day precision (`<1h`, `Nh`).
- **A silent degradation path.** Five summariser failure modes returned correct output while logging nothing, so a failing model looked identical to a channel of short pins.

Two limits worth stating rather than implying: the digest tracks a pin's **source text**, not the summariser's version, so a summariser improvement does not re-summarise pins already stored — right for cost, but it means an old line only heals when its message is edited. And the store was cleared by hand mid-session to exercise the fixes, so the live run does **not** cover "an existing store survives a version upgrade"; that rests on AC11's unit coverage.

## Unresolved

Grouped by whether this change introduced the thing, and by how much is actually known about it.

### Introduced by this change, measured

| Item | Certainty |
|---|---|
| **A summariser improvement does not reach pins already stored.** The digest tracks a pin's source text, so an improved summary only lands when the message itself is edited. Deliberate — the alternative re-bills every pin in every channel on every deploy. | Certain — observed live |
| **Six long decoy pins can force a seventh onto the verbatim path** by exhausting `MAX_MODEL_CALLS_PER_SCAN` inside one TTL window. `source="verbatim"` discloses *that* a line is unsummarised, but not *why*. | Measured by QA |
| **Fullwidth `＜＞` pass through unescaped.** Structurally inert in XML — they cannot form a tag — but the reader is a language model. | Measured by QA; judged not worth widening the escape |
| **A steady-state scan is not free.** One `pins.list` plus one `users.info` per distinct author or pinner, per TTL window per channel. Only the model call is cached. | Measured; documented |
| **`by` and `pinned_by` render self-chosen Slack display names.** A member can name themselves after a colleague. The block's note says names prove nothing; nothing prevents it. | Certain — inherent to Slack, shared with every other surface in this system |

### Introduced by this change, not established either way

| Item | Certainty |
|---|---|
| **"The task spawns normally" on a failed `pins.list` was never executed end to end.** The repo has no test that drives `spawn.ts` prompt assembly at all. What *is* proven is that the builder always resolves in both failure states, that `ensureChannelPins` swallows even a store-layer throw, and that real tasks ran normally throughout the live session. | Suspected-safe, unproven directly |
| **Digest reuse is strong evidence, not proof.** Byte-identical summaries after a genuine re-scan; nothing logs "reused" versus "summarised", so a model returning the same sentence twice would look the same. A debug line would settle it. | Measured, inferential |

### Pre-existing, surfaced by this work

| Item | Certainty |
|---|---|
| **Issue #265** — context edited mid-task is not re-shown to an already-briefed agent until its next wake. Pins inherit it; this change does not fix it. | Certain — known bug |
| **`docs/guides/local-development.md`'s permission list is a 4-scope subset** omitting a dozen scopes including `pins:read`. Left alone: it is the minimal set for a working local bot, and pins is optional. | Certain — deliberate |
| **`isExternalUser` fails open when the home team id is unknown.** Shared with the canvas gate and every other caller. | Certain — pre-existing |

### Process notes

| Item | Certainty |
|---|---|
| **`AC2` was reworded after sign-off.** As signed it said a failed `pins.list` emits no block; `AC20` requires retained pins to keep rendering through exactly that failure. The code satisfies `AC20`; the criterion now distinguishes a channel with nothing stored from one holding pins. Flagging because the operator signed the earlier wording. | Certain |
| **The first QA cycle ran four runners concurrently against one checkout**, and their scratch test files briefly contaminated `AC17`. They cleaned up; `AC17` was re-run against a confirmed-clean tree. My dispatch error, not a code issue. | Certain |
| **`AC10` was attacked three times** and the first two attacks broke the mechanism then in place. The third found no escape in 351 fresh renders. | Certain |
| **`AC18` shipped as a waiver and has since been verified live**, so the manifest row below understates it. The manifest is left as computed rather than hand-edited. | Certain |

## Deployment & follow-ups

**`pins:read` has now been granted and the app reinstalled** in the Sweatcoin workspace, which is what made the live run above possible. For any other workspace, this ships dormant and does nothing until the same steps are taken — `pins.list` returns `missing_scope`, the feature no-ops, and no task is affected.

To turn it on:

1. Re-import `slack-manifest.yaml` at api.slack.com.
2. Reinstall the app to the workspace. **This rotates the bot token.**
3. Swap `SLACK_BOT_TOKEN` and restart. The restart is required, not optional — the first `missing_scope` latches a process-level flag that suppresses every later `pins.list`, so a running instance never picks the scope up on its own.
4. Then the `AC18` check: pin a message in a channel Archie is in, mention Archie there, and ask it what is pinned and to open one of them.

Step 2 is the one with real risk — the old token stops working, so Archie is down between the reinstall and the redeploy.


egorkhmelev:
Follow-up filed: #270 — give the PM a `read_channel_pins` exploration tool.

This PR adds the **push** direction only: a standing index for channels the task is already linked to. #270 covers the **pull** direction — asking what any reachable channel has pinned — which needs no new scope and would reuse `listChannelPins` and the existing accessible-channel gate. It deliberately differs on three points (no summariser, a lower-but-labelled trust bar since a tool result is not the system prompt, and a `limit` like its sibling tools); the reasoning is in the issue.

PR #280, opened 2026-08-17
fix(slack): let watched-channel bot posts reach trigger dispatch

hardworker:
## The problem

A channel-message trigger could never fire on a report posted by a bot through a legacy incoming webhook. Two independent causes, both upstream of the trigger itself:

- **The event was discarded before triggers were consulted.** Those posts arrive with `subtype: bot_message`, which the inbound subtype gate dropped outright.
- **The filter tested an empty string.** Their `text` is empty and all content lives in attachments, which Bolt strips at the event boundary — so the `contains` filter matched against `""`.

## The fix

- `bot_message` now passes the gate, but **only as a watched top-level channel post** — never a thread reply, never a DM. A channel with no trigger watching it behaves exactly as before, so the blast radius is the set of channels that already have a trigger.
- The `contains` filter matches the re-fetched root message rendered through `renderMessageForContext` — the same helper `read_channel_history` already uses for attachment-only messages. Root-only on purpose: rendering replies too would let a human quoting the phrase fire the trigger.

Deliberately unchanged: the matcher stays inline, `FireContext.text` stays unread, and no agent gains a tool or permission.

## Verified live, both ways

Run against real instances fed by real Slack events, with a throwaway trigger watching #archie-test-channel for "Expired Feature Flags Report". The probe went through a genuine Slack incoming webhook, reproducing the report's shape exactly — `subtype: bot_message`, empty `text`, phrase only in `attachments[0].text` — with a third-party `bot_id` (`B0BQKB5G0KF`), so it was never a self-authored post.

| Build | Message | Result |
|---|---|---|
| base `544a538` | [webhook probe](https://sweatcoin.slack.com/archives/C0BL50N2600/p1786963661555699) | **nothing** — no task, no fire, no log line |
| base `544a538` | [control](https://sweatcoin.slack.com/archives/C0BL50N2600/p1786963556094269) — phrase in `text` | **fired** → [`PROBE OK`](https://sweatcoin.slack.com/archives/C0BL50N2600/p1786963569596199?thread_ts=1786963556.094269) in-thread |
| this PR | [same webhook probe](https://sweatcoin.slack.com/archives/C0BL50N2600/p1786970098895989) | **fired** → [`PROBE OK`](https://sweatcoin.slack.com/archives/C0BL50N2600/p1786970116511159?thread_ts=1786970098.895989) in the probe's own thread, 18s later |

`544a538` is the commit production runs, and its control fired seconds before the probe was ignored — so the silence was the subtype gate, not a dead pipeline. On this PR the reply lands in the webhook message's own thread, which is exactly the path the real report needs for its file attachment.

## Also verified

- Against the four real messages of an actual report: all four were dropped before, all four now reach dispatch, and only the header matches the filter — one fire per report rather than one per chunk.
- `npm run typecheck` and `npm run build` clean; new routing test covers all four arms of the bot case. The one suite failure, `TaskList.test.tsx`, fails identically on unmodified `main`.

## Before merge

Production pins its image tag in the systemd unit on `archie-hq0`, so merging alone changes nothing until someone bumps it and restarts.

### DIFF
--- a/src/connectors/slack/__tests__/task-routing.test.ts
+++ b/src/connectors/slack/__tests__/task-routing.test.ts
@@ -55,3 +55,37 @@ describe('group-DM (G…) routing parity', () => {
     ).toBe(true);
   });
 });
+
+describe('bot_message routing (legacy-webhook bot posts)', () => {
+  // Bots posting through a legacy webhook carry `subtype: 'bot_message'`, which the
+  // subtype gate used to drop before trigger dispatch was ever consulted. They are
+  // forwarded only as a watched top-level channel post — the case the pre-existing
+  // "trigger is watching" test above missed, because it used a subtype-less human post.
+  const botPost = { type: 'message', subtype: 'bot_message', channel: 'C0XYZ', ts: '1' };
+
+  it('forwards a bot post when a trigger is watching the channel', () => {
+    expect(shouldForwardMessageEvent(botPost, () => true)).toBe(true);
+  });
+
+  it('ignores a bot post when no trigger is watching', () => {
+    expect(shouldForwardMessageEvent(botPost, () => false)).toBe(false);
+  });
+
+  it('ignores a bot thread reply even when a trigger is watching', () => {
+    expect(
+      shouldForwardMessageEvent({ ...botPost, ts: '2', thread_ts: '1' }, () => true),
+    ).toBe(false);
+  });
+
+  it('ignores a bot message in a DM', () => {
+    expect(
+      shouldForwardMessageEvent({ ...botPost, channel: 'D0ABC' }, () => true),
+    ).toBe(false);
+  });
+
+  it('still drops other subtypes, watched or not', () => {
+    expect(
+      shouldForwardMessageEvent({ ...botPost, subtype: 'message_changed' }, () => true),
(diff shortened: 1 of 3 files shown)


egorkhmelev:
Superseded by #291, which landed both halves of this in a more general form.

The subtype allowlist this PR extends with `bot_message` was inverted into a **denylist** (`MESSAGE_NOISE_SUBTYPES`), and the routing question was split into `mayWakeTask` / `mayReachTriggers` — so an app post reaches trigger dispatch without spending a PM turn. That also covers the case this PR would still have missed: an app calling `chat.postMessage` with a bot token arrives with `bot_id` and *no* subtype at all, so keying on `subtype === 'bot_message'` catches only the incoming-webhook shape.

The `text` argument threaded through `dispatchChannelMessageTriggers` here solved the same problem #291 solved with `rawMessageBody(event.raw, channelId)` inside the dispatch function, so a `contains` filter now matches attachment-only bot reports. The diagnosis in this PR — that Bolt leaves `event.text` empty for a webhook bot while the fetched root carries the content — was right, and it's what both fixes are built on.

Closing as superseded. #298 then built on that render path for trigger delivery.

PR #281, opened 2026-08-17
Unify core skill mounting behind one manifest and one ordered list

egorkhmelev:
## What & why

Which agents receive one of archie-hq's own skills was decided by which directory a file sat in. `AgentDef` carried two skill-path fields — `skillsPath`, a plugin's own `skills/` dir, and `coreSkillsPath`, archie-hq's built-in `skills/` — and `coreSkillsPath` was set on exactly one def, the PM. So a core skill's audience was implicit in its location, and giving one a second audience would have meant a second directory.

This replaces both fields with one ordered list, `skillPaths?: string[]`, built from a central manifest that declares which core skills each agent **track** mounts. Plugin skills keep expanding from the plugin's own directory and stay first in the order, so shadowing is unchanged. Workspace setup becomes a plain symlinker over that one list. The manifest encodes today's mapping exactly — all four core skills on the `pm` track, nothing on the others — so **nothing changes behaviourally**, and a boot warning names any core skill no track mounts, so forgetting the manifest is announced instead of silent.

Adding an audience is now a line in `CORE_SKILL_MOUNTS` rather than a new directory and a new field.

## How it works

`src/agents/core-skills.ts` is the one place that knows about core skills. It owns the (private) core-skills directory constant, the `CORE_SKILL_MOUNTS` manifest keyed `pm` / `repo` / `plain`, and `resolveSkillPaths(track, pluginSkillsPath?)`, which returns the ordered, basename-deduplicated list of absolute skill directories. Plugin entries come first, which *is* the shadowing rule. `setupAgentWorkspace` then just symlinks the list.

**The track is not stored on the def.** There is no `track` field in this codebase — the only classification is the two predicates `isRepoAgent` and `isPmAgent`, with a plain plugin agent as the negation of both. Rather than add a field a fifth construction site could set wrongly, each site passes the literal it already branches on: `scanAgentDefs` passes `'repo'` or `'plain'` per its existing `if (agent.repo)` fork, `buildPmDef` passes `'pm'`, and `synthesizeDynamicAgentDef` passes `'repo'` because a PM-spawned dynamic agent is a repo agent by construction. That one falls out at today's behaviour — the `repo` track is empty, so it mounts nothing, exactly as before.

Expansion moves from spawn time to registry-scan time, because a list that mixed "a directory to expand" with "a single skill directory" would not be a plain symlinker. Two consequences, both deliberate and both now documented: `scanAgentDefs` performs roughly a dozen small `readdirSync` calls where it previously performed one `existsSync`, and it runs on every `Task.create` / `Task.get`; and `skillPaths` is a snapshot, so a *removed* skill is handled by a re-check at mount time while an *added* one waits for the next scan.

That re-check deserves its own note, because it is the one place review found a real regression. Trusting the snapshot without it meant `symlink()` could be called for a source that had since disappeared — and `symlink(2)` does not validate its target, so it would create a **dangling** link. A dangling link at the mount target then wedges that agent permanently, since the target guard is `existsSync`, which follows the link, reports false, and lets `symlink()` throw `EEXIST` out of a path that does not catch it. The old loop got this for free by enumerating the directory at spawn time. One `existsSync` on the source restores it.

**Deliberately out of scope.** `LoadedPlugin.skillsPath` is a different field on a different type and is untouched. No new audience is added. The sandbox `/app` `denyRead` boundary is unchanged. The pre-existing dangling-link `EEXIST` fatality is filed separately rather than fixed here, so that this change stays provably behaviour-preserving — see Unresolved.

## Verification

Gate: `npm run typecheck`, `npm run build`, `npm test` all exit 0 — **80 test files / 1126 tests passing**, up from 78 / 1108 on `main`. `npm run lint` is excluded because it is broken repo-wide (`sh: eslint: command not found`) and CI does not run it either.

| id | criterion | method | status | where it ran | evidence |
|---|---|---|---|---|---|
| AC1 | `skillsPath`/`coreSkillsPath` are gone from `AgentDef` and every def carries one ordered list | structural | **verified** | `command` | Field-declaration grep empty (exit 1); `skillPaths?: string[]` present at `src/types/agent.ts:215`; typecheck exit 0. |
| AC2 | The mounted skill set for all 13 defs is identical to the pre-change baseline — PM 26 (22 plugin + 4 core), every other def 1 source | integration | **verified** | `suite` | `core-skills-equivalence.test.ts` ran (5 passed, 0 skipped) against the real plugins clone. Corroborated three independent ways: a captured pre-change vs post-change per-def artifact that diffs **byte-identical**; a reviewer transcribing the *old* two-source mount logic and diffing it against the new field across all 13 defs, 0 mismatches; and off-disk counts (22 plugin dirs + 4 core = 26) reproduced by two agents. |
| AC3 | A plugin skill of the same name shadows a core skill | unit | **verified** | `suite` | `core-skills.test.ts` collision case. Falsifiable: disabling the dedupe reddens it; reordering the source blocks reddens 3 tests. |
| AC4 | A symlink-vendored skill mounts; a dangling symlink is skipped | unit | **verified** | `suite` | `core-skills.test.ts` symlink case. Falsifiable: dropping the classification reddens 5 tests; accepting every symlink reddens the dangling assertion. |
| AC5 | A core skill no track mounts produces `logger.warn('system', …)` naming it; otherwise no warning | unit | **verified** | `declared-harness` | Computation covered both directions in unit tests. **Emission observed live**: planting `skills/zz-qa-unmounted-probe` and restarting produced exactly one line — ``[system] Core skill "zz-qa-unmounted-probe" is mounted by no agent track — add it to CORE_SKILL_MOUNTS…`` — with the PM's `skills:` line correctly excluding it. Removing the probe drove the count back to zero. |
| AC6 | Only the manifest module builds core-skill paths | structural | **verified** | `command` | `CORE_SKILLS_DIR` appears as *code* only in `src/agents/core-skills.ts`; zero code references to either removed field anywhere in `src/`. See the deviation below. |
| AC7 | A PM-spawned dynamic agent resolves through the manifest on the `repo` track and mounts nothing | unit | **verified** | `command` | Probe on the real `synthesizeDynamicAgentDef` returns `skillPaths: []`. Proven to flow from the manifest, not a hard-coded empty list: setting the `repo` track to `['triggers']` moves the synthesized list. |
| AC8 | The boot banner's `skills:` line derives solely from the one list and is byte-identical to today's output | unit | **verified** | `suite` | Single derivation confirmed — `index.ts` calls the exported `mountedSkillNames` and no second implementation exists. The running container's actual boot-log `skills:` line matched the predicted 26 sorted names character for character. |
| AC9 | typecheck + build + suite all pass, ≥1108 tests | structural | **verified** | `command` | Exit 0 / 0 / 0; `Test Files 80 passed (80)`, `Tests 1126 passed (1126)`, zero skipped. |
| AC10 | The two architecture docs describe the manifest, and the infra comments no longer name a deleted field | manual | **waived** | `none` | Accepted as `manual` at sign-off. **Reviewer checklist:** confirm `docs/architecture/plugin-system.md` (the new "Core skills, and which tracks mount them" section) and `docs/architecture/agents.md:74`, plus the reworded comments in `Dockerfile.prod` and `docker-compose.yml`. A review step can confirm edits that exist but cannot force them to exist, which is why the targets are named here. |

Beyond the criteria, two live checks ran against the booted harness: `tools/e2e/egress-check.ts` passed with both directions asserted (its header requires a run after any `src/agents/spawn.ts` change, and this changes it), and a **real spawned agent's workspace held 26 symlinks with 0 dangling**, the four core ones resolving to `/app/skills/<name>` and readable through the link, with the agent's own reply listing those 26 names in order. That last one matters because it exercises `setupAgentWorkspace`, which has no unit coverage.

Every new test was mutation-checked — the behaviour each one guards was broken and the test confirmed to go red. Two that could *not* fail were found and fixed before merge.

### liveVerification

> No criterion required a live instance, so nothing was expected to run against one.

That sentence is computed from the criteria, and it is accurate: no AC was signed as `live-e2e`. Live verification nonetheless happened and is stronger than the plan required — AC5's emission and the two extra checks above all ran against the repo's declared `archie-e2e` harness, booted from this branch with container sources hash-verified against the working tree. The instance was torn down afterwards and the credentials its setup copied in were removed.

## Unresolved

| # | Item | Introduced by this change? | How I know |
|---|---|---|---|
| 1 | A dangling symlink **already** at a mount target is still fatal: the target guard is `existsSync`, which follows the link, so `symlink()` throws `EEXIST` uncaught when a stale link coexists with a live source. Fix is `lstatSync(target)`, unlink-and-relink, or catching `EEXIST`. | **No — pre-existing.** This change strictly improves it by removing the only path that *creates* new dangling links. | Measured. A reviewer ran a verbatim replica of the `origin/main` loop and got the identical `EEXIST`. Filed as its own issue rather than fixed here, to keep this change provably behaviour-preserving. |
| 2 | The one-line source re-check in `setupAgentWorkspace` has no unit test, because the function is module-private and the repo has no `spawn.ts` test file. | **New** (the line is new). | Known by construction. Exercised live instead — see the spawned-agent workspace check above. Closing it properly means extracting the mount loop behind an injectable seam. |
| 3 | `scanAgentDefs` now does ~a dozen small `readdirSync`/`statSync` calls where it did one `existsSync`, on a path that runs per `Task.create`/`Task.get`. | **New**, and inherent to making the symlinker plain. | Read off the code and called out in the design so it is judged rather than discovered. Not measured under load. |
| 4 | `core-skills-equivalence.test.ts` **skips entirely in CI** — `PLUGINS_DIR` lives under the gitignored `workdir/`, and CI checks out only archie-hq. A green CI is not evidence that mount-set equivalence holds. | **New** (the file is new). | Measured — run with an empty workdir it reports 5 skipped, exit 0. Disclosed in the file's own header. The CI-visible regression net is the fixture-based `core-skills.test.ts`. |
| 5 | That file's basename-uniqueness assertion cannot fail against today's tree, since no plugin skill collides with a core name. | **New.** | Measured — removing the dedupe entirely leaves it green. Documented in place; the falsifiable proof of the dedupe is the fixture test. |
| 6 | Dropping `.sort()` in `findUnmountedCoreSkills` is caught by no test, because `readdir` happens to return the fixtures pre-sorted. Cosmetic — it only orders warning lines. | **New.** | Measured by mutation. |
| 7 | The `secrets/` directory itself matches no `.gitignore` rule. Every file in it today is covered by a specific pattern, but a future credential dropped there under another name would be untracked and committable. | **No — pre-existing.** | Verified with `git check-ignore`. Out of scope; a `secrets/*` plus `!secrets/.gitkeep` rule would close it. |
| 8 | AC6 deviates from the criterion as literally worded: both removed field names survive as **prose inside comments** in the two new test files, and those files recompute the repo's `skills/` path for their own assertions. No code outside the manifest module builds a core-skill path. | **New** (the comments are new). | Measured by grep, and recorded rather than rounded to a pass. |
| 9 | AC2's strongest artifact — the byte-identical pre/post capture — was produced by the driver, not by a blind runner. | n/a (process). | Recorded as corroboration, not independent evidence. The independent confirmations are a reviewer's transcription of the old logic and two agents' off-disk counts. |
| 10 | AC8's byte-identity was established against an independent recomputation and the live boot log, not against a rebuilt pre-refactor binary. | n/a (process). | Stated plainly rather than implied. |

Deviations from the verification plan, for completeness: AC2 and AC6 ran extra commands beyond the plan; AC5 was re-run on the live harness after its unit evidence was judged to cover only the computation; AC8's counts came from a probe; AC9 ran as three commands rather than one `&&` chain.


egorkhmelev:
Follow-up for the deferred fix, filed rather than folded in so this PR's behaviour-preservation claim stayed provable: #282 (pre-existing dangling-symlink `EEXIST` wedge). This PR removes the only path that could create a new dangling link, so it makes that issue strictly better without closing it.

egorkhmelev:
### CodeQL: 3 new high alerts — same three sinks as `main`, relocated

The `CodeQL` check is red with *"3 new alerts including 3 high severity security vulnerabilities"*. All three are `js/path-injection` in `src/agents/spawn.ts`, and they are the **same three operations `main` already carries**, re-reported because this PR rewrote those lines:

| `main` | this PR | operation |
|---|---|---|
| `spawn.ts:141` | `spawn.ts:128` | `existsSync(target)` |
| `spawn.ts:142` | `spawn.ts:129` | `mkdir(agentSkillsDir, …)` |
| `spawn.ts:143` | `spawn.ts:130` | `symlink(…, target)` |

Same rule, same sinks, same taint source (a skill directory name reaching an `fs` call), one for one. The equivalents on `main` are open alerts #67, #68, #72 and #73 — `ref: refs/heads/main`, so they predate this branch. No new sink and no new taint source is introduced; if anything the target is now derived through `basename()`, where the old loop used a bare `readdir` entry name.

The other three checks are green: **Typecheck, build & test** (80 files / 1126 tests), **Analyze (javascript-typescript)**, and **Secret scan (gitleaks)**.

I have deliberately not touched this. Hardening `spawn.ts`'s path handling means deciding what a valid skill name is and validating it at the boundary, which is a real change with its own design question and its own tests — not something to bolt onto a behaviour-preserving refactor. Happy to either dismiss these three as pre-existing-and-relocated, or file it as a follow-up issue alongside #282; say which you prefer.

egorkhmelev:
**Correction and resolution on the CodeQL note above.**

Two fixes to that comment. The mount-block equivalents on `main` are **#67, #68 and #73** — I wrote #72, which is actually `spawn.ts:165` (the `settings.json` write), a different sink this PR does not touch. And the check is **informational**: `main` has no branch protection, so no required status check exists and the red X does not block merge.

For the record, since the comment ended with an open question: **decision is to leave all three alerts open, with no dismissal and no follow-up issue.** They are `main`'s #67/#68/#73 relocated by this rewrite, not new findings.

Worth writing down for whoever picks this up later. Neither the PR alert nor its `main` equivalent reports any dataflow path, so the rule is firing on the sink shape rather than a traced request-to-sink route. The skill-name component is passed through `basename()` and cannot traverse; `taskId` comes from `generateTaskId()` (date plus random, `src/tasks/persistence.ts:37`) and `POST /api/tasks` destructures only `{ message }`. The one genuinely request-controlled path component in this class is `req.params.id` used as a task id at `src/connectors/api/routes.ts:126`, `:199` and `:223` — pre-existing, upstream of these lines, and where a real fix would go: validating it against the known task-id format at the API boundary would close the whole class at once.

PR #288, opened 2026-08-18
fix(sandbox): let node-gyp find its devdir and the Node headers

egorkhmelev:
Fixes two of the three things blocking native addon builds in the sandbox. Follow-up to #283, from a gap that PR's live run surfaced.

**Native builds still do not work after this** — a third cause is outside its scope, see the end.

## 1. devdir pointed at an unwritable `$HOME`

node-gyp keeps the Node headers it compiles native addons against in a "devdir". With none configured it falls back to `envPaths('node-gyp').cache` — `$HOME/.cache/node-gyp` (`node-gyp/bin/node-gyp.js`). That path does not exist in the image and `$HOME` is not in `allowWrite`, so the mkdir failed with `ENOENT` (not `EROFS`, because the missing parent is `.cache`, not the leaf).

Same class as the npm/yarn `EROFS` that `2b23da0` fixed. `npm_config_devdir` points it at the shared cache instead.

Why that env var: node-gyp scans `npm_config_*` from the environment and maps them onto its own options (`node-gyp/lib/node-gyp.js`), so it works even though npm is not setting it. The prefix is deprecated in npm v11 for `npm_package_config_node_gyp_*`, which deliberately **outranks** it — so setting the old one leaves a repo free to override the devdir from its own `package.json`. A test pins that choice.

## 2. nodejs.org was not on the egress allowlist

With the devdir fixed the download still failed:

```
gyp http GET https://nodejs.org/download/release/v24.19.0/node-v24.19.0-headers.tar.gz
gyp http 403 ...
gyp ERR! configure error
```

Added `nodejs.org` to `TRUSTED_PACKAGE_REGISTRY_DOMAINS`. That constant is documented as deliberately minimal, so the reasoning is on the record:

- **Exfiltration** — a static download host: no upload endpoint, no API, no user content echoed back. Weaker than `registry.npmjs.org` already on the list, where `npm publish` would carry data out.
- **Code execution** — no meaningful addition; an agent can already install and run arbitrary packages from the registry.
- **Integrity** — node-gyp verifies downloads against `SHASUMS256.txt`. That defeats tampering in transit, though not a compromise of the host itself — the same trust already extended to the registry.
- **Trust tier** — OpenJS Foundation, HTTPS, comparable to the registries. Not a general-purpose host.

Considered and rejected: seeding the headers into the image at build time. It avoids touching the allowlist, but only covers the image's own Node version — a repo needing a different ABI would 403 again.

## Why nobody noticed any of this

npm and yarn treat a failed *optional* postinstall as a warning and still exit 0. On `sweatcoin-frontend`, `@newrelic/native-metrics` and `@contrast/fn-inspect` silently fell back to JS. A repo with a *required* native dependency would have failed outright.

## Verification

`npm run typecheck` clean; suite **1125 passed, 5 skipped**.

Live on `sweatcoin-frontend` (yarn 4.5.3, edit mode approved) via `archie-e2e`:

| Claim | Evidence |
| --- | --- |
| devdir resolves to the shared cache | `devdir=/workdir/caches/node-gyp` |
| Headers now download | `gyp http 200`; **no 403 in any log** (grepped yarn log + both build logs) |
| Headers land on disk | devdir holds `24.19.0/{include,installVersion}`, 2726 header files incl. `include/node/node.h` |
| Egress boundary intact | egress check passes both directions; `npm install: ok` |

One trap worth recording: `ls -d node_modules/<pkg>/build` **succeeds** for both packages even on failure, because those dirs hold only gyp-generated files (`Makefile`, `config.gypi`, `*.target.mk`) and no compiled `.node`. Presence of `build/` is not evidence of a successful build.

## Still broken: no C toolchain in the image

```
gyp ERR! stack Error: not found: make
```

`make`, `gcc`, `g++` and `cc` are all absent from the image (`python3` is present). So configure now succeeds and the compile step fails.

Fixing that means adding a build toolchain to the image — an image-size and runtime-attack-surface decision rather than a sandbox one, so it is deliberately not bundled here. Worth its own change.

### DIFF
--- a/src/agents/__tests__/sandbox.test.ts
+++ b/src/agents/__tests__/sandbox.test.ts
@@ -34,10 +34,10 @@ describe('buildManagedNetworkPolicy', () => {
       ...base,
       allowedNetworkDomains: [...TRUSTED_PACKAGE_REGISTRY_DOMAINS],
     });
-    expect(policy.sandbox.network.allowedDomains).toEqual([
-      'registry.npmjs.org',
-      'registry.yarnpkg.com',
-    ]);
+    // Compared against the input, not a copy of it: this test is about the policy
+    // passing through what it is handed. The allowlist's actual contents are
+    // pinned separately, so growing the constant does not fail this.
+    expect(policy.sandbox.network.allowedDomains).toEqual([...TRUSTED_PACKAGE_REGISTRY_DOMAINS]);
   });
 
   it('always sets allowManagedDomainsOnly — the allowlist is not enforced without it', () => {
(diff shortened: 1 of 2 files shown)


egorkhmelev:
Closing — not proceeding for now. The change works (headers download, devdir populated, verified live) but does not deliver working native builds on its own: the image has no C toolchain, so the compile step still fails with `not found: make`. Findings above stay accurate if anyone picks this up later.

PR #293, opened 2026-08-18
fix(triggers): make a scheduled delivery open and link the task's thread

egorkhmelev:
**Stacked on #292** — base is `fix/trigger-message-context`, because it restructures the same fire path. Merge #292 first and this retargets to `main` cleanly.

## The problem, measured live

A schedule fire has no thread to reply in, so the PM delivered through `post_to_channel`, which is deliberately fire-and-forget. Before this change, on a live instance:

- fired task after a successful delivery: `channels: {}`, `default_channel: None`
- the tool's own result: `Message posted to C0BL50N2600 (new thread ts: 1787084582.563519). Not linked to this task.`
- a human reply in that thread: `[System] Created task task-20260818-2025-mxwntp` — a **stranger task**. Fresh PM, no trigger id, no `triggered_by`, no access to the trigger's directory. The thread got linked *there* instead.
- the PM also had to invent the mandate `post_to_channel` demands (`"Trigger instruction from the user who set up trigger trg-…"`), because there is no human in that task's thread to quote.

## The change

`fireTrigger` records `metadata.delivery_target` for a channel-bound schedule fire. `Task.postToUser` acts on it: posts top-level there, stores the returned ts as a linked channel, makes it the default, clears the target so later messages thread under it. `findTaskByThread` matches on `thread_id`, so replies come back to the task that posted. The tool's `New channel linked: …` return path was already written for exactly this and previously had no way to fire.

**`post_to_channel` is untouched** — still the tool for a channel this task does not live in, still links nothing. The two jobs are now separate paths instead of one path doing both badly.

## The load-bearing part was the PM's context, not the prompt

First live attempt: target recorded correctly, and the PM still delivered nothing. Its own reasoning, verbatim:

> "Task metadata says 'Channel(s): none — there is nowhere to reply in this task; finish with report_completion() (no message).'"
> "There is no channel linked to this task, so `post_to_user` has no destination."

That context line was telling it to give up. With a delivery target it now names the destination and says the first post opens the thread.

## Live verification, after

Same probe trigger, same channel:

| | before | after |
|---|---|---|
| tool the PM chose | `post_to_channel` (after ToolSearch) | `post_to_user`, no target |
| tool result | "Not linked to this task." | "New channel linked: slack:C0BL50N2600:1787086486.442039" |
| fired task metadata | `channels: {}` / `default_channel: None` | `default_channel: slack:C0BL50N2600:1787086486.442039`, `delivery_target: null` |
| human reply in the thread | `Created task task-20260818-2025-mxwntp` | **`Resumed pm-agent for task task-20260818-2054-wt8m06`** — the fired task, no new task |

The reply was injected as a signed event through `/webhooks/slack` (bot tokens cannot post as a human). The bot's root message and the thread were real; the reply itself was synthetic, so the **routing decision** is what this verifies — not the ingest of the reply's text, which `Task.append` skips because the message does not exist in Slack.

## Also

- `post_to_user` and `report_completion(message)` no longer refuse when `channels` is empty but a delivery target is set
- 5 new unit tests in `src/tasks/__tests__/delivery-target.test.ts`: top-level post (no `threadTs`), the linked channel's exact shape, `findTaskByThread` resolving the posted ts, target cleared so the second message threads, no-ts leaves it sent-but-unlinked rather than inventing an id, and the target being ignored once a default channel exists
- `buildTriggerSeed` drops the channel-id delivery line for channel-bound schedule fires — delivery is `post_to_user` like everywhere else. The user-bound (DM) case keeps its line and is unchanged: the id worth storing is the `D…` conversation while the message is addressed to a `U…`, so linking it needs a `conversations.open` this does not do
- `npm run build` clean, `npx vitest run` 1168 passed

### DIFF
--- a/docs/architecture/triggers.md
+++ b/docs/architecture/triggers.md
@@ -72,8 +72,18 @@ A one-off schedule auto-pauses after it fires (its condition is consumed). **Res
 `fireTrigger(trigger, context)` (`src/system/trigger-scheduler.ts`) is shared by the scheduler (schedule context) and the Slack dispatch hook (message context):
 
 1. Create a fresh task; set `metadata.triggered_by = trigger.id`.
-2. For a message-context fire, ingest the triggering thread with `Task.append` — knowledge log, linked channel, default channel, no post; for a schedule fire, the spawned PM opens the destination itself.
+2. Wire delivery. A message fire ingests the triggering thread with `Task.append` — knowledge log, linked channel, default channel, no post. A channel-bound schedule fire records `metadata.delivery_target`, and the PM's first user-facing message opens a thread there and links it.
 3. Seed the PM with `AGENT_PROMPTS.triggered(...)` and let it do the work.
+
+**A scheduled delivery becomes the task's thread.** A schedule fire has no thread to reply in, so `fireTrigger` records `metadata.delivery_target` — the bound channel — and `Task.postToUser` does the rest: the first user-facing message is posted top-level there, the ts that comes back is stored as a linked channel, that channel becomes the default, and the target is cleared so every later message threads under it.
+
+Two things follow. The PM delivers with plain `post_to_user` and no argument, exactly as in any other task, instead of being handed a channel id to route by hand. And **a human replying to a scheduled post comes back to the task that posted it** — `findTaskByThread` matches on `thread_id`, which is now that ts.
+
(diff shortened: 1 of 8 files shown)


egorkhmelev:
Closing — the operator is redoing this change from scratch in a separate session. Branch `fix/trigger-delivery-links-thread` is left in place for its live evidence and tests.

PR #311, opened 2026-08-31
perf(pm): cut time-to-first-word 20% by cutting thinking, not round trips

egorkhmelev:
Archie takes about 14 seconds to say its first word. This makes it take about 11, and explains why.

**Land with https://github.com/sweatco/archie-plugins/pull/200.** The measured win is the pair. This half did not reach significance alone.

## What the time is actually made of

Measured on a live instance from this branch, 303 runs across two campaigns:

| phase | mean | share |
|---|---|---|
| think / ttft | 11.2s | 67% |
| streaming | 2.6s | 16% |
| dispatch | 2.5s | 15% |
| tool execution | 0.6s | 3% |

**83% of the wait is the model emitting tokens. 3% is doing anything.**

That single line decided everything else. Five candidate tweaks were implemented behind independent flags and measured separately, and the two that cut thinking won while the two that removed a round trip did not — because round trips were never the expensive part.

| tweak | mechanism moved? | latency | verdict |
|---|---|---|---|
| PM effort high→medium | — | −1.7s, p=0.0047 | **ship** (replicated: −2.2s, p=0.0103 in campaign 1) |
| lean reasoning section | — | −0.8s, p=0.13 | **ship as a pair** — did not replicate alone |
| **both together** | — | **−3.2s (20%), p=0.0001** | **ship** |
| `alwaysLoad` on hot MCP tools | ToolSearch 1.00 → 0.12/sample | −1.6s, p=0.10 | dropped — fired cleanly, bought nothing |
| inject knowledge.log into the turn | log fetches 1.20 → 0.00/sample | −1.0s, p=0.25 | dropped — round-trip cost *rose* 2.8s → 4.3s |
| strip malformed `pages` arg | 0 occurrences on CLI | n/a | unresolved — no target on the tested channel |

Output tokens corroborate independently: 846 → 552 for effort, → 649 for the prompt. The ranking of token reduction matches the ranking of wins exactly.

## What changed

**`prompts/pm-agent.md`** — the `<situation_analysis>` scaffold (nine sections, a worked example, ~170 lines, "thoroughness is more important than brevity here") becomes five questions. That instruction predates extended thinking; it now runs *inside* it, so the same deliberation is paid for twice — and it is invisible, since no PM session on disk has an assistant text block over 20 characters. Every load-bearing check survives: the thread-conduct gate, stop-means-mute, don't-conclude-while-outstanding, turn-ending discipline, load-the-skill-first.

**`tools/speed/`** — the suite all of this came from. Pure functions over records Archie already writes, so any campaign can be re-read offline. Its waterfall is what produced the table above.

## Why you can believe the numbers

- **Comparisons run on ranks** (Mann–Whitney U, tie-corrected). Latency is heavy-tailed; one outlier moves a mean enough to invent an effect.
- **A comparison across differing models or serving tiers is refused**, with a non-zero exit, rather than rendered.
- **A null arm calibrated the noise.** The `pages` tweak turned out to have no target, making it an accidental placebo. At n=20 it read −1.5s and beat baseline on 4/5 cases; at n=25 it converged to −0.1s, p=0.559. Nothing not clearly better than that is distinguishable from doing nothing.
- **Each arm declares the counter it should move**, so "did nothing" and "had no target" are different verdicts.

## Does it break anything

Four adversarial cases probe the rules a shorter prompt might drop: being told to stop (mute and stay silent — it genuinely calls `mute_channel`, 15/16), an injection payload embedded in pasted user text, a question nothing in context can answer, and two instructions that cannot both hold.

**8 of 9 cases pass 4/4 on every arm.** Totals: baseline 94%, effort 97%, prompt 97%, combined 94%. The one discriminating case fails equally on the *unmodified* prompt, so it is a pre-existing weakness, not a regression.

That is evidence of no harm on what the suite can reach — not proof. See below.

## Limits, stated plainly

- **CLI channel only.** All 303 runs. Slack tasks are more expensive per turn and load skills ~3× as often; historical Slack p50 was 24.6s against this CLI baseline of 14.1s. The mechanism (less thinking per turn) is channel-independent; the magnitude is unverified.
- **No group DM, no multi-agent, no edit mode.** `hops = 0.00` across all 303 runs — delegation never fired once, and the 27.6s-per-hop cost is untouched by any of this.
- **The `pages` tweak is unresolved, not dead.** All 27 malformed `Read` calls in the historical data came from Slack; zero from CLI. It has no target on the channel that was tested.
- **~2.3s of per-turn engine dispatch is untouched** by everything here, and is not spawn cost — a second turn with the agent already running still pays 2.30s against 2.59s cold. That is the next thing worth looking at.


egorkhmelev:
Closing: the prompt half is not worth the trade. The nine-section scaffold stays.

The measurement harness in `tools/speed/` is not part of that decision and will be re-opened on its own branch — it is what produced the phase breakdown (83% of a turn's wall-clock is the model emitting tokens, 3% is tool execution) and the finding that ~2.3s of every turn is CLI subprocess cold start, paid on every user message because `complete()` tears the agent down.

The effort change (sweatco/archie-plugins#200) is superseded by a better idea: change effort dynamically rather than permanently.

PR #321, opened 2026-09-02
test(voice): the case harness and test suite for the voice participant

egorkhmelev:
The measurement half of the Zoom voice participant, split out so the agent code it exercises can be read on its own. **[The agent code is stacked on top of this](https://github.com/sweatco/archie-hq/pulls) — review that one for the behaviour; this one is the instrument, not the thing measured.**

The split exists because the harness is 73% of the diff. Reading the agent's design through 22k lines of fixtures and detectors was the problem.

## `tools/voice-cases/`

A case harness that runs the real prompts against the real provider and grades what comes back. Ten of its files run under `npm test` at zero cost against canned transports; the drivers themselves bill an API and are run by hand.

Every one of those ten exists because an instrument nobody tested had already produced a confidently wrong conclusion here:

- Three detectors silently never fired on Russian input, because JS's `\b` and `\w` are ASCII-only and never bound a Cyrillic letter. One was still described in a comment as *working* while an ordinary Russian phrasing produced no hit at all.
- A driver called its own builder with two arguments and no arm wiring, so no row had ever carried the context block it claimed to measure — every stored row had run the wrong branch of the prompt.
- A grader that was never called scored everything clean, which no unit test of the grader could see.

What the suites cover: single-turn defect cases with per-family verdicts derived rather than hand-declared, multi-turn chains, the triage gate under two context arms, long-transcript fixtures whose inertness is asserted rather than assumed, retry and rate-limit behaviour on a driven clock, and sample accounting that reports rows lost to a rate limit as a banner rather than quietly shrinking a denominator.

## `src/**/__tests__/`

Unit and integration tests for the connector, the meeting state machine, the transports, and the task binding. The ones worth knowing about assert things a reader would otherwise delete as redundant:

- Controls that must pass both before *and* after a fix, so an over-broad change cannot slip through.
- A test that awaits a full event-loop turn rather than draining microtasks, because Node emits `unhandledRejection` only after microtasks drain — drain alone and the test asserts nothing.
- One piece of deliberately degraded behaviour, pinned so nobody fixes it away.

## Notes for review

**This branch does not pass on its own.** These tests import modules that arrive in the stacked branch. The stack is green as a whole: 131 files, 2,554 tests, plus 4 from `main`'s recent Slack test.

Comments across both branches were cut to what a reader could not derive from the code — 877k characters to 297k. What survives is provider facts, non-local couplings, and decisions that look like bugs.


egorkhmelev:
Superseded by #323. See #322 for the reasoning — the tests belonged with the code, and the stack order had to put the code first.

PR #322, opened 2026-09-02
feat(voice): Archie joins a Zoom meeting as a participant

egorkhmelev:
Archie dials into a meeting as a virtual colleague rather than a notetaker: it listens, works out when it is being addressed, decides where the answer has to come from, and speaks.

**Stacked on [#321](https://github.com/sweatco/archie-hq/pull/321), which carries the case harness and tests.** This branch is the agent itself — 39 files, ~8.5k lines — so the diff here is the code to read. The harness was 73% of the combined diff and was drowning it.

## How a turn works

Recall puts a bot in the room and streams per-participant audio to us; Deepgram Flux turns that into utterances. **Archie speaks only when no participant holds an open turn** — the room's silence is what gives it the floor, and the owe flag is conversation state alongside the transcript.

Each turn runs three model calls, all prepared from one shared context builder and differing only in job:

1. **Addressing gate** — is this for Archie?
2. **Triage gate** — where does the answer live: the room, outside, already-asked, or nowhere this team's records could settle. The gate structurally cannot see its own verdict.
3. **Speaking call** — what to say.

A verdict of `outside` carries one sentence Archie says aloud while it goes and finds out, so the room isn't left in silence through a full model call.

## What it can do

Speak, write to the meeting chat, ask the PM a question that leaves the room, and leave. One question can be outstanding at a time. Everything said is recorded to a transcript beside the meeting, and the PM writes the meeting up afterwards in the thread that asked Archie to join.

Deepgram Aura has no Russian voice, so Soniox carries both — one voice in either language, routed on the presence of Cyrillic rather than its proportion.

## Things that look like bugs and are not

These are the comments worth reading, because the code cannot say them:

- **The capability summariser is deliberately un-awaited on the join path.** Awaiting it would hold a live meeting behind a provider's latency, and a hung one would stop the join outright. A turn taken before it lands simply has no capability block.
- **A `.catch` exists solely to stop a process-level crash.** There is no global `unhandledRejection` handler anywhere in `src/`, so one uncaught rejection would take down every other task and meeting.
- **`MeetingHost` methods are synchronous void and swallow their own failures**, for the same reason — the loop calling them is an end-of-turn reaction with about three hundred milliseconds to answer.

## Provider behaviour recorded where it bites

- `language_hint` must be a **repeated query key**. The comma-joined form is silently accepted while corrupting both language detection and transcript punctuation.
- Deepgram validates parameter *names* strictly but not values, so a typo'd hint is only catchable via a field in the response.
- Soniox synthesises nothing until `text_end`, and silently deletes square-bracket audio tags.

## Review notes

Comments were cut to the minimum a reader could not derive from the code — 877k characters to 297k across both branches — in two passes. The first deleted comments that shouldn't exist; the second cut the wording of what was left. Surviving comments are provider facts, non-local couplings, and the deliberate-looking-like-a-bug decisions above.

Two things known and deliberately not addressed here: the 3-hour transcript window is arbitrary rather than calibrated (its original justification was arithmetic against a context ceiling that turned out to be half the real one, so it now has roughly double the headroom anyone believed), and the ~40k-token three-hour worst case has no test — everything measured is the routine one-hour shape.

Green as a stack: 131 files, 2,554 tests, typecheck clean.


egorkhmelev:
Superseded by #324. The split was wrong: it cut the tests away from the code they test, which made the base PR unmergeable on its own. Redone as #323 (agent code plus its own tests, green standalone) with #324 stacked on top carrying only the `tools/voice-cases/` harness — which has to come second because it imports `src/voice/comprehension.ts`.

PR #323, opened 2026-09-02
feat(voice): Archie joins a Zoom meeting as a participant

egorkhmelev:
Archie dials into a meeting as a virtual colleague rather than a notetaker: it listens, works out when it is being addressed, decides where the answer has to come from, and speaks. This is the agent and its own tests, and it stands alone — green and mergeable by itself.

The `tools/voice-cases/` case harness is a separate PR stacked on top of this one. It imports `src/voice/comprehension.ts`, so it cannot land first.

**Reading the diff:** 66 files, ~21k lines, of which the code is ~8.5k and its tests ~12.5k. The code worth reading is concentrated in `src/voice/comprehension.ts` (1,470), `src/voice/meeting.ts` (1,368), `src/voice/deepgram.ts` (713), `src/connectors/recall/audio-out.ts` (706), `src/connectors/recall/index.ts` (683) and `src/voice/soniox.ts` (510).

## How a turn works

Recall puts a bot in the room and streams per-participant audio to us; Deepgram Flux turns that into utterances. **Archie speaks only when no participant holds an open turn** — the room's silence is what gives it the floor, and the owe flag is conversation state alongside the transcript.

Each turn runs three model calls, all prepared from one shared context builder and differing only in job:

1. **Addressing gate** — is this for Archie?
2. **Triage gate** — where does the answer live: the room, outside, already asked, or nowhere this team's records could settle. The gate structurally cannot see its own verdict.
3. **Speaking call** — what to say.

A verdict of `outside` carries one sentence Archie says aloud while it goes and finds out, so the room isn't left in silence through a full model call.

## What it can do

Speak, write to the meeting chat, ask the PM a question that leaves the room, and leave. One question can be outstanding at a time. Everything said is recorded to a transcript beside the meeting, and the PM writes the meeting up afterwards in the thread that asked Archie to join.

Deepgram Aura has no Russian voice, so Soniox carries both — one voice in either language, routed on the presence of Cyrillic rather than its proportion.

## Things that look like bugs and are not

These are the comments worth reading, because the code cannot say them:

- **The capability summariser is deliberately un-awaited on the join path.** Awaiting it would hold a live meeting behind a provider's latency, and a hung one would stop the join outright. A turn taken before it lands simply has no capability block.
- **A `.catch` exists solely to stop a process-level crash.** There is no global `unhandledRejection` handler anywhere in `src/`, so one uncaught rejection would take down every other task and meeting.
- **`MeetingHost` methods are synchronous void and swallow their own failures**, for the same reason — the loop calling them is an end-of-turn reaction with about three hundred milliseconds to answer.

## Provider behaviour recorded where it bites

- `language_hint` must be a **repeated query key**. The comma-joined form is silently accepted while corrupting both language detection and transcript punctuation.
- Deepgram validates parameter *names* strictly but not values, so a typo'd hint is only catchable via a field in the response.
- Soniox synthesises nothing until `text_end`, and silently deletes square-bracket audio tags.

## Review notes

Comments were cut to the minimum a reader could not derive from the code — 877k characters to 297k across this and the stacked PR — in two passes: the first deleted comments that shouldn't exist, the second cut the wording of what was left. What survives is provider facts, non-local couplings, and the decisions above.

Two things known and deliberately not addressed here. The 3-hour transcript window is arbitrary rather than calibrated: its original justification was arithmetic against a context ceiling that turned out to be half the real one, so it now has roughly double the headroom anyone believed. And the ~40k-token three-hour worst case has no test — everything measured is the routine one-hour shape.

**Verification:** standalone on this branch, 122 test files / 2,159 tests pass, `tsc --noEmit` clean.

## Simplification pass (two follow-up commits)

**One vendor per job** (`c8fbbf35`). Deepgram is Flux (STT) only; Soniox is the only synthesizer; Cerebras `gemma-4-31b` is the only comprehension model. `speech.ts`, the Aura implementation, the Anthropic provider and both `ARCHIE_VOICE_*_PROVIDER` selectors are gone, along with every "this synthesizer cannot report that" degradation branch. `SONIOX_API_KEY` and `CEREBRAS_API_KEY` are now required for the connector to mount. −1,252 / +266.

**One module, one teardown signal** (`eaf55207`). `src/connectors/recall/` is merged into `src/voice/`; `VoiceTransport.sendChat`, `AudioSink.setEngaged` and `AudioSink.playedBytes` are required, and the branches for their absence are gone; `join_recall_meeting`/`leave_recall_meeting` are built over closures on the mount's own start and stop instead of a starter/stopper function-pointer registry. The 45 s empty-room grace, the socket-close-then-ask-Recall path and the 3-minute idle sweep are replaced by Recall's own `automatic_leave.everyone_left_timeout` (set to 60 s — the default is 2 s, which reads a Zoom rejoin as an empty room) plus a 30 s poll of the bot's `status_changes` that ends the meeting on `call_ended`/`done`/`fatal`. −1,362 / +730.

Net: the voice agent's non-test source is roughly 2,600 lines smaller. 122 files / 2,125 tests pass standalone. Not yet exercised against a live Recall bot — the `automatic_leave` body shape comes from Recall's API reference and the poll is tested against a mocked ledger.

### Further passes (same day)

- **One speaking call** (`68a16c6e`) — triage gate, `<situation>` block, spoken preamble and the `ARCHIE_VOICE_PROMPT_PLACEMENT` experiment removed; every turn loses one serial model round-trip. Addressing gate (all three tiers) and the capabilities summary unchanged.
- **Native reasoning** (`8c2d2621`) — the speaking call uses Cerebras' `reasoning_effort` on `gemma-4-31b`; the prompted `<think>` tag parser is gone; the prompt keeps its thinking guidance and drops only the tag syntax. Measured on the voice-case harness, 3 decorrelated reps per arm: overall defect rate 4.9% → 4.9%, quality 45/48 → 45/48, first spoken token median 306 → 697 ms.
- **Credentials-only config** (`8eb92ffb`) — `DEEPGRAM_HOST`, `SONIOX_HOST`, `SONIOX_VOICE`, `ARCHIE_VOICE_LANGUAGE_HINTS`, `ARCHIE_VOICE_BOT_NAME` become constants; one flat `VoiceConfig`; the two PM wake-ups move to `prompts/voice-wakeup-*.md`.
- **One record per meeting** (`324efda5`) — `transcript.log`/`exchange.log`/`chat.log`/`metadata.json`/`capabilities.json` and the activation log become one `recall/<sessionId>/meeting.jsonl`; platform and title are captured by the status poll; no teardown fetch.
- **The page reports what it played** (`9cb4a7b1`) — the output hub's server-side playback estimation and suppression window are replaced by the worklet reporting rendered samples; the confirmed-heard transcript on barge-in is now a measurement.
- **One process-level rejection handler** (`f2915c07`) — ten local `safe*` wrappers and the "never throws" contracts go; `docs/proposals/zoom-voice-participant.md` (implemented, misfiled) deleted.
- **Rewrite** (`997642da`) — `meeting.ts` and `comprehension.ts` rewritten against the unchanged suite: 1,101 → 912 and 1,011 → 746 lines, 24 → 12 per-meeting state variables.

Not yet exercised in a live meeting: the `automatic_leave` create-body field, the 30 s status poll, and the page's `played` reporting.


src/memory/lifecycle.ts
egorkhmelev, line 1: why did you do this change?

src/agents/spawn.ts
```diff
@@ -420,6 +413,12 @@ Shared folder: ${sharedPath} [READ-ONLY]
     mcpServers['comms-tools'] = createCommsMcpServer(agent, task);
     mcpServers['orchestration-tools'] = createOrchestrationMcpServer(agent, task);
     mcpServers['scheduling-tools'] = createSchedulingMcpServer(agent, task);
+    // Whatever a mounted connector registered for the PM (e.g. Recall's
+    // `recall-tools`, holding `join_recall_meeting`) — never named here, so a
+    // deployment where nothing registered anything hands the PM nothing extra.
+    for (const [name, factory] of getRegisteredConnectorPmTools()) {
```
egorkhmelev: why recall get this, but not slack for example? aren't we overcomplicating this?

src/agents/spawn.ts
```diff
@@ -346,15 +348,6 @@ export async function spawnAgent(agent: Agent, task: Task): Promise<void> {
     systemPrompt = await generatePMPrompt(task);
 
     const channelEntries = Object.entries(metadata.channels);
-    const renderChannel = (id: string, ch: typeof metadata.channels[string]): string => {
```
egorkhmelev, line 349: is there a reason you refactored this into separate file?

prompts/voice-speaking.md
egorkhmelev, line 1: thing I don't understand is that why there is no additional context in this prompt? maybe its good idea to add here everything for transparency as variables, so we can see what comes first/last and what exactly in the context of speaking agent just by reading its prompt

src/tasks/channel-delivery.ts
egorkhmelev, line 1: feels like overcomplication, defend it

src/tasks/persistence.ts
```diff
@@ -155,6 +156,42 @@ export function getArtifactsPath(taskId: string): string {
...
+ *
+ * One folder per MEETING, not one per task: `sessionId` is the same id that
+ * keys the meeting's `recall:<sessionId>` channel key (see `recallChannelKey`
+ * in `src/voice/task-binding.ts`), so the folder and the channel key are
+ * built from the same value and can never drift apart — no separate lookup
+ * table maps one to the other. A task that hosts several meetings over its
+ * life gets one sibling folder per meeting here, rather than every meeting's
+ * speech landing in one shared file behind an artificial boundary line.
+ */
+export function getMeetingPath(taskId: string, sessionId: string): string {
```
egorkhmelev: why is it here at all?

src/tasks/persistence.ts
```diff
@@ -474,6 +511,87 @@ export async function appendCliMessage(
...
+ * Mirrors `sanitizeForLog` in `src/voice/meeting.ts`, and applied again here,
+ * at the point of writing, so a consult question — model output derived from
+ * an open microphone, which this codebase can verify no better than a
+ * self-reported name (see `prompts/voice-wakeup-question.md`) — gets the same
+ * guarantee regardless of what a caller passes in. Deliberately narrow: only
+ * control characters and the two Unicode line separators some consumers treat
+ * the same way are touched, so ordinary punctuation, non-Latin scripts and
+ * emoji in real text survive untouched.
+ */
+function sanitizeTranscriptField(value: string): string {
```
egorkhmelev, line 493: this and whole thing below, its place is with recall connector

src/types/task.ts
```diff
@@ -88,7 +88,7 @@ export interface SlackThread {
 
 // ---- Channel types (replace slack_threads) ----
 
-export type ChannelType = 'slack' | 'github' | 'cli';
```
egorkhmelev, line 91: again, why? no one asked you to drop github

src/types/task.ts
```diff
@@ -134,7 +127,40 @@ export interface CliChannel extends ChannelBase {
...
+ * `src/voice/channel-delivery.ts`.
+ *
+ * The record itself is permanent, exactly like a `SlackChannel`: linked and
+ * `ended: false` when the meeting starts, flipped to `ended: true` when it
+ * stops, never removed. That is what lets a post to this channel — live,
+ * ended, or left stale by a crash — always reach the Recall connector's own
+ * deliverer rather than fall through to generic task code that has no idea
+ * what a meeting is. See `src/tasks/channel-delivery.ts`.
+ */
+export interface RecallChannel extends ChannelBase {
```
egorkhmelev, line 161: why its in task?

egorkhmelev (commented):
first batch of review

src/memory/lifecycle.ts
egorkhmelev: It should not have changed at all — the GitHub-channel removal reached this file too. Reverted in a3577a8c; `src/memory/lifecycle.ts` is byte-identical to `main` again.

src/types/task.ts
egorkhmelev: Nothing needed it gone and it went anyway. Restored in a3577a8c, with `recall` added beside GitHub in the union rather than in its place.

src/agents/spawn.ts
egorkhmelev: No good one. `renderChannel` is back in `spawn.ts` beside the strings it produces (72f0ff06), and it kept a test per channel kind — the registry test was the only thing asserting `Meeting (live)`/`Meeting (ended)`.

src/agents/spawn.ts
egorkhmelev: You are right, and the registry is gone (72f0ff06): recall is now named in `postToUser` beside Slack. Three delivery branches to teach instead of a registry to register with, which is the cheaper of the two.

src/tasks/channel-delivery.ts
egorkhmelev: Not defending it — deleted in 72f0ff06. Slack was already named directly in both places the registry served, so a second mechanism for the same job bought only a layer to read through.

src/tasks/persistence.ts
egorkhmelev: It should not have been. `getMeetingPath` moved to `src/connectors/recall/meeting-record.ts` in 72f0ff06, and its CodeQL path-traversal alerts moved with it (the sanitizer and the all-dots arm came along).

src/tasks/persistence.ts
egorkhmelev: Agreed — `getMeetingRecordPath` and `appendMeetingRow` are in `src/connectors/recall/meeting-record.ts` now (72f0ff06). `notifyMeetingEnded` takes the path instead of deriving it, so `src/voice/` still imports nothing from `src/connectors/`.

src/types/task.ts
egorkhmelev: Because it is a record in the task own `metadata.json`, exactly like `SlackChannel` and `CliChannel` — the task is what owns the list of places it can be reached. Left as it is; the stale pointer to the deleted `channel-delivery.ts` in that comment is fixed.

prompts/voice-speaking.md
egorkhmelev: Done in bbfdb898: the whole context block is now `prompts/voice-speaking-context.md` with named placeholders, so the order the model reads it in is visible in the file rather than assembled by string concatenation in `comprehension.ts`. A test reads the placeholders out of the template and requires the builder to fill every one.

PR #333, opened 2026-09-13
chore(migrate): one summary line for dropped pre-v30 repository entries

egorkhmelev:
## Summary

- Production startup recovery reads every task's metadata, and 22 tasks from March 2026 still carry the pre-v30 `repositories` map shape (short keys like `"backend"`) that `migrateRepositoriesShape` in `src/tasks/task.ts` drops. Each dropped key logged its own warn line, so a single boot produced 44 lines of noise — and the same lines repeated on every later load of those tasks.
- The drop itself is correct: those entries' `github` identifier can no longer be resolved (the agent registry that used to resolve it is gone), the migration is in-memory only, and the repo is simply re-mounted on demand when the task needs it. Only the log volume was the problem.
- Replaced the per-key warn with a module-level counter/task-id tracker and a single deferred summary: the first drop across the process logs one warn line with the running entry/task counts (e.g. `"2 legacy pre-v30 repository entries dropped in memory across 1 tasks (first: task-a); they predate github-keyed repositories and are re-mounted on demand"`), and another summary line only fires if the set of affected tasks grows by ten since the last one. Everything else logs at debug instead of disappearing.
- Added `src/tasks/__tests__/repositories-migration.test.ts` coverage asserting the per-task/per-key warn is gone (one warn call covers a task with two dropped keys, not two), that the summary line carries the counts, and that a follow-up summary appears once the affected-task set grows by ten.

## Test plan

- [x] `npm run typecheck`
- [x] `npx vitest run src/tasks/__tests__/repositories-migration.test.ts` (10/10 passing)
- [x] `npm test` (107 files / 1617 tests passing)


egorkhmelev:
Superseded by #334: the repeated lines came from webhook lookups migrating every candidate task, which no longer happens. The one line per key for a task actually picked up is a useful signal and disappears once activation persists the flat shape.

